PR #47, opened 2021-10-04
Rename Vocab/Che to Vocab/Red-Hat

themr0c:
Rename CheDocs Vocabulary into RedHat.
Fix inconsistencies. 
Fix broken rules.
Fix broken content (errors only).


themr0c:
This is fixing stuff causing errors in https://github.com/rolfedh/studious-fortnight/pull/36

PR #53, opened 2021-10-06
Fixes: symbolic link creation step.

themr0c:
Fixes: symbolic link creation step.

Can someone test it on macOS for me?


themr0c:
This pull request is obslete. The content has moved to the Antora website (with the fix).

PR #57, opened 2021-10-19
One unique redistributable style: *RedHat*

themr0c:
To redistribute the style in a form consumable by the GitHub Action, we must create a ZIP archive containing one single directory with all rules, including the Spelling rules.

* Move all rules to one unique style: *RedHat*.
* Reimplement Vocabulary into a *Spelling* rule that can be distributed with the RedHat Style. Vocabulary should be used by users to modulate the style. Update the documentation accrodingly.
* Clean up the Vale configuration file: use only the unified RedHat style. Remove the Vale style as we have now a proper Spelling rule in the RedHat style.
* Create a Vale configuration file suitable for the GitHub Action.
* Create a `release.yml` workflow to release a `RedHat.zip` file on push to main.
* Write tests for some rules: ProductNames, Spelling, url, yaml.
* Fix vale errors on random files.


themr0c:
Linting with Vale failure is expected, as the required ZIP and configuration file are not yet available.

PR #71, opened 2021-10-26
New words in the Spelling rule

themr0c:
* Modify the Spelling rule
* Fix the procedure: Extending the Spelling rule
* Ignore some files in `tools/report_vale_alerts.sh`


themr0c:
10 days without a comment, I guess it's OK to merge :)

PR #96, opened 2021-11-26
Create Contractions.yml

themr0c:
IBM style guide says: Contractions are permitted in all types of content.

Borrowed from https://github.com/testthedocs/PlainLanguage


themr0c:

PR #97, opened 2021-11-26
Create ComplexWords.yml

themr0c:
Borrowed from https://github.com/testthedocs/PlainLanguage/blob/master/ComplexWords.yml


themr0c:

PR #102, opened 2021-11-26
Create Conjunctions.yml

themr0c:
Do not overuse beginning sentences with '%s'


themr0c:

PR #103, opened 2021-11-26
Create Cliches.yml

themr0c:
source: https://github.com/errata-ai/proselint/blob/master/proselint/Cliches.yml


themr0c:

PR #113, opened 2021-12-02
fix: don't raise an error for 'shell'

themr0c:
A "shell" is a software application (for example, /bin/bash or /bin/sh) that provides an interface to a computer. Do not use this term to describe the prompt where you type commands.


themr0c:
fixes https://github.com/Vale-at-Red-Hat/vale-at-red-hat/issues/88

PR #132, opened 2021-12-15
fix: remove warning on "additional"

themr0c:
Don't raise a warning for "Additional resources"


.vale/styles/RedHat/ComplexWords.yml
```diff
@@ -18,7 +18,6 @@ swap:
   accurate: "'right' or 'exact'"
   acquiesce: "'agree'"
   acquire: "'get' or 'buy'"
-  additional: "'more' or 'extra'"
```
gaurav-nelson, line 21: @themr0c If we are removing a legitimate word, shouldn't we add it back in some other rule that can differentiate between `additional` and `additional resources`?

Issue #137, opened 2021-12-18
Display warnings and errors by default, not suggestions


rolfedh:
**Describe the bug**
Usability issue: I believe the current configuration of `.vale.ini` has `MinAlertLevel = suggestion`.

Unfortunately: 
- Running vale on a typical file generates an overwhelming number of suggestions. 
- The suggestions are too repetitive.
- Some of the suggestions are misguided. 

I believe this experience demotivates users from trusting and using vale. 

I believe the solution is to increase the default value of `MinAlertLevel = warning` in `.vale.ini` and to ensure that the warnings and errors are not misguided.

Examples:

With `MinAlertLevel = suggestion`, you get:
```
rdlugyhe@rdlugyhe openshift-docs (RHDEVDOCS-3570 *<) $ vale --minAlertLevel=suggestion  modules/images-other-jenkins-config-kubernetes.adoc

 modules/images-other-jenkins-config-kubernetes.adoc
 6:3      suggestion  'Configuring the Jenkins        RedHat.Headings       
                      Kubernetes plug-in' should use                        
                      sentence-style capitalization.                        
 8:209    warning     Consider using 'many' instead   RedHat.ComplexWords   
                      of 'multiple'                                         
 10:90    suggestion  Do not use 'as' to mean         RedHat.As             
                      'because' or 'while'.                                 
 12:1     suggestion  Try to keep sentences less      RedHat.SentenceLength 
                      than 25 words.                                        
 12:1     suggestion  Verify your use of 'Both' with  RedHat.Usage          
                      the word usage guidelines.                            
 12:70    suggestion  Do not use 'as' to mean         RedHat.As             
                      'because' or 'while'.                                 
 12:187   suggestion  Try to keep sentences less      RedHat.SentenceLength 
                      than 25 words.                                        
 12:254   suggestion  'be applied' is passive voice.  RedHat.PassiveVoice   
                      Consider rewording in active                          
                      voice. You can use the passive                        
                      voice in prerequisites                                
 12:337   suggestion  'be run' is passive voice.      RedHat.PassiveVoice   
                      Consider rewording in active                          
                      voice. You can use the passive                        
                      voice in prerequisites                                
 12:366   suggestion  'is applied' is passive voice.  RedHat.PassiveVoice   
                      Consider rewording in active                          
                      voice. You can use the passive                        
                      voice in prerequisites                                
 12:428   suggestion  Verify your use of              RedHat.Usage          
[...]

PR #138, opened 2021-12-18
Display warnings and errors by default, not suggestions #137

rolfedh:
Display warnings and errors by default, not suggestions #137


themr0c:
I apologize, I didn't meant to close the issue before we reach consensus. I made a typo in a commit.

themr0c (changes requested):
I don't agree to remove suggestions as a default.

Suggestions are meaningful in general. And they are meaningful in the #137 example.

I agree to remove suggestions as a temporary action when onboarding a project, as documented there: https://vale-at-red-hat.github.io/vale-at-red-hat/docs/end-user-guide/defining-a-vale-onboarding-strategy/

Opening specific issues on the rule that you consider misguided would be a far more constructive approach

michaelryanpeter (commented):
I find suggestions helpful, as both a reviewer and a writer. 

What if we update the docs with the following:

- How to use use the `--minAlertLevel` flag
- A brief explanation of why suggestions are useful, and a fewcontexts where one might want to turn off suggestions
- Maybe we could have a small shell script to turn off suggestions?

WDYT?

themr0c:
@michaelryanpeter 

> How to use use the --minAlertLevel flag

See: https://redhat-documentation.github.io/vale-at-red-hat/docs/end-user-guide/defining-a-vale-onboarding-strategy/#_focus_on_errors

> A brief explanation of why suggestions are useful, and a fewcontexts where one might want to turn off suggestions

See: https://redhat-documentation.github.io/vale-at-red-hat/docs/end-user-guide/understanding-vale-output/#understanding-vale-alert-levels_docs
and https://redhat-documentation.github.io/vale-at-red-hat/docs/end-user-guide/defining-a-vale-onboarding-strategy/#_focus_on_errors

> Maybe we could have a small shell script to turn off suggestions?

Wouldn't is be more complex than learning to use the --minAlertLevel flag?

michaelryanpeter:
Thank you for pointing these out, @themr0c. I have spent more time in the documentation over the past few weeks. I think that you are right, having writers copy and paste something into their `.bashrc` is probably more complex than learning how to use the flags.

I have an idea for discussion. It is a tangent from @rolfedh's original issue, but it might help. I will open an idea/issue and refer back to here.

PR #147, opened 2022-01-10
Update ssl-tls.yml

themr0c:
https://redhat-documentation.github.io/supplementary-style-guide/#ssl-tls

 SSL/TLS (noun)

Description: SSL/TLS refers to the Secure Socket Layer protocol (SSL) and its successor, the Transport Layer Security protocol (TLS). As both of these protocols are frequently called "SSL", always use "SSL/TLS" to avoid confusion.

Use it: yes

Incorrect forms: SSL, TLS, TLS/SSL


themr0c:
TODO: make sure we match "SSL", "TLS", and we do not match "SSL/TLS".
TODO: add test files.

Issue #129, opened 2021-12-13
To indicate that a user should input text, use "enter" rather than "type" or "input".


themr0c:
Implement https://redhat-documentation.github.io/supplementary-style-guide/#text-entry

level: suggestion

PR #150, opened 2022-01-14
addressing issue #129: indicate that a user should input text, use "enter" rather than "type" or "input"

kquinn1204:
HI @themr0c this is my attempt to address https://github.com/redhat-documentation/vale-at-red-hat/issues/129. I am not sure if it is the correct syntax

### DIFF
--- a/.vale/styles/RedHat/text-enter.yml
+++ b/.vale/styles/RedHat/text-enter.yml
@@ -1,0 +1,10 @@
+extends: substitution
+message: "Use '%s' rather than '%s'."
+link: https://redhat-documentation.github.io/supplementary-style-guide/#text-entry
+ignorecase: true
+level: error
+action:
+  name: replace
+source: Red Hat
+swap:
+  type|input: enter


themr0c:
Looks good. Linting with Vale timeout when there is no content is a known issue.

themr0c:
Ideally, add a fixtures directory including test files. Get inspiration from: https://github.com/redhat-documentation/vale-at-red-hat/tree/main/.vale/fixtures/RedHat/ConfigMap

themr0c:
* For consistency with the other rules, please use CamelCase and for the file name use a word that won't trigger a Vale alert. For instance, rename the file: Enter.yml

* Add a fixtures directory with test files.

* Add an entry for the reference guide: rebase on `main`, run the (new) `./tools/update-reference-guide.sh` script, edit the resulting AsciiDoc file generated in the Reference guidr.

themr0c:
We can do it even more simply now.

I created the pull request as a demonstration on how to add en entry in an existing rule:

https://github.com/redhat-documentation/vale-at-red-hat/pull/168/files

PR #153, opened 2022-01-19
fix #114 - no error for "terminal"

### LINKED ISSUE #114: 'Terminal emulation' is legit
"Terminal emulation" refers to making a computer respond like a particular type of terminal. Terminal emulation programs allow you to access a mainframe computer or bulletin board service with a personal computer.

Use it: yes

https://redhat-documentation.github.io/supplementary-style-guide/#_t

However the shell-prompt rule is raisinn an error: https://github.com/Vale-at-Red-Hat/vale-at-red-hat/blob/main/.vale/styles/RedHat/shell-prompt.yml

themr0c:


themr0c:
The  Linting with Vale / Linting with Vale error is expected.

Issue #145, opened 2022-01-05
vale errors on hostname, TLS, terminal, OS, vscode


max-cx:
**Describe the bug**
The following words or expressions trigger illegitimate vale alerts:

**1.** 
We need `hostname` as per IMB Style on p. 292:
![image](https://user-images.githubusercontent.com/78564304/148204387-3437ea35-c36d-4d29-9536-8316b9326739.png)
**I suggest downgrading this rule to a warning to avoid massive fixes of `host name` to older docs and to allow for use of `hostname` in new docs.**

**2.**
The following screenshot is self-explanatory--the lintered text is **`SSL TLS`**:
![image](https://user-images.githubusercontent.com/78564304/148206878-ffeb120c-0d36-4989-a613-f310948f9351.png)
How it happened:
At first I had just `TLS` in the text: this was blocked by Vale as an error.
So I changed it to `SSL TLS`: :smile: now I got two errors, one for `SSL` (Vale doesn't see TLS with it) and one for `TLS` (Vale doesn't see SSL with it).
In my docs, only TLS is mentioned. If I have to add SSL, that may be technically incorrect:
https://www.internetsociety.org/deploy360/tls/basics/
https://www.ssl.com/faqs/faq-what-is-ssl/
**I suggest removing this rule also because SSL and TLS don't seem to be the same thing and in many situations only one of them should be mentioned.**

**3.**
![image](https://user-images.githubusercontent.com/78564304/148204632-f7a229e3-80ab-4710-9e5e-a0234bae0ff2.png)
The above rule prevents me from using phrases like "in the terminal window". I can live with having to use `shell prompt`, if I'm forced to do so by Vale, to describe how to enter a command; however, in other situations, like when asking the user to (for example) verify the output of a command in the terminal window, the meaning doesn't work with `shell prompt` (or even with `command line`). So based on my experience, `shell prompt` and `terminal` are not interchangeable (and by extension also `terminal window` because then I can't write even `terminal window` as Vale will block it as an error due to `terminal` in it).
Please also note that IMB Style does not prohibit use of the word `terminal`, as can be seen on p. 347: 
![image](https://user-images.githubusercontent.com/78564304/148205013-a537063b-d2b0-4b6c-9fa3-aca624983687.png)
**I suggest downgrading this rule to a warning as explained.**

**4.**
[...]

PR #155, opened 2022-01-19
Max cx patch 1

max-cx:
Downgrading severity on `ProductNames.yml` and `operating-system.yml` as a solution to points 4 and 5 in https://github.com/redhat-documentation/vale-at-red-hat/issues/145

Affecting all entries in [ProductNames.yml](https://github.com/redhat-documentation/vale-at-red-hat/blob/19fee48d8655ea6e05057179d60b26798edb4537/.vale/styles/RedHat/ProductNames.yml) because the same problem can happen to any of them (occurrences in third-party links, etc.).

Please check if the following lines need to be updated in the edited files:
```
action:
  name: replace
```

### DIFF
--- a/.vale/styles/RedHat/ProductNames.yml
+++ b/.vale/styles/RedHat/ProductNames.yml
@@ -1,7 +1,7 @@
 extends: substitution
 message: "Use '%s' rather than '%s'."
 ignorecase: false
-level: error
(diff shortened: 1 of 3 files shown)


max-cx:
Discussed this with Fabrice. Wrong solution.

PR #213, opened 2022-05-09
Updating ISG links

aireilly:
Since all Red Hatters now have access to ISG with RH logins, this PR updates the links to the ISG online in the VRH docs.


antora.yml
```diff
@@ -27,7 +27,7 @@ asciidoc:
     # for the project
     context: docs
     ibmsg: IBM Style Guide
-    ibmsg-url: https://www.pearson.com/us/higher-education/program/De-Respinis-IBM-Style-Guide-The-Conventions-for-Writers-and-Editors/PGM270594.html?tab=contents
+    ibmsg-url: https://www.ibm.com/docs/en/ibm-style
```
themr0c, line 30: Meaning we replace a public resource by a private one? I don't have access to this page.

aireilly: Updated, leaving the original links, and also adding the new ones.

Issue #209, opened 2022-04-25
'Machine' incorrectly flagged with a style guide suggestion in instances of Java Virtual Machine (JVM)


michelle-purcell:
**Describe the bug**

The word 'Machine' is incorrectly flagged in instances of Java Virtual Machine (JVM):

![image](https://user-images.githubusercontent.com/92924207/165125568-7e4b42be-b3b9-4b3e-a191-cce790f7c3d4.png)

The suggestion is to 
`
vale: Verify your use of 'Machine' with the word usage guidelines. (RedHat.Usage)
`

**To Reproduce**

1. Add this word or expression: Java Virtual Machine 
2. Vale reports an illegitimate alert: error, warning, suggestion: 
* This word or expression triggers an illegitimate vale alert of type suggestion. 'Machine'

**Expected behavior**
Java Virtual Machine is a proper noun. It's an official product name. When your content includes 'Java Virtual Machine' this creates annoying suggestions, which can create unnecessary churn.

This relates to the following issue: https://github.com/redhat-documentation/vale-at-red-hat/issues/189
While issue 189 requests Red Hat product names are recognized, ideally commonly used third-party products and services should also be detected and excluded from style guidance.

aireilly:
Fixed in https://github.com/redhat-documentation/vale-at-red-hat/pull/226

PR #226, opened 2022-06-14
Fix issue #209: Do not flag valid uses of "machine"

jhradilek:
As per the style guide, both "virtual machine" and "machine learning" are permitted. This fix ensures that these valid uses of the word "machine", including variations such as "Java Virtual Machine", are no longer flagged.

### DIFF
--- a/.vale/styles/RedHat/Usage.yml
+++ b/.vale/styles/RedHat/Usage.yml
@@ -96,7 +96,7 @@ tokens:
   - localize
   - look and feel
   - look-and-feel
-  - machine
+  - "(?<!virtual )machine(?! learning)"
   - master and slave
   - migrate
   - native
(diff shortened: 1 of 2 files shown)


.vale/styles/RedHat/Usage.yml
```diff
@@ -96,7 +96,7 @@ tokens:
   - localize
   - look and feel
   - look-and-feel
-  - machine
+  - "(?<!virtual )machine(?! learning)"
```
aireilly, line 99: Is there a benefit to using this regex instead of:

```
- virtual machine
- machine learning
```

aireilly: Oops. My bad. You are correct :)

PR #227, opened 2022-06-14
Do not flag "backtrace" and "traceback"

jhradilek:
Both "backtrace" and "traceback" are valid terms in computing and used frequently when documenting debugging procedures. They are not listed flagged as incorrect in any of our style guides so I do not see a reason why they should be flagged by Vale as incorrect.


aireilly:
LGTM

PR #228, opened 2022-06-14
Do not flag "such as"

jhradilek:
While the style guide does not explicitly states its opinion on using "such as" in product documentation, it uses it excessively throughout the Word usage section. As it is a common phrase, not ignoring it caused the term suggestion for "as" to generate too many misguided suggestions.


aireilly:
LGTM

PR #229, opened 2022-06-14
Do not flag "DVD" as an acronym to define

jhradilek:
Similarly to CD, DVD should not be flagged for acronym expansion.


aireilly:
LGTM

PR #230, opened 2022-06-14
Do not flag /var when used as an obvious path

jhradilek:
The supplementary style guide defines VAR as an acronym for "value-added reseller" but unfortunately the existing definition also incorrectly reports all occurrences of "/var" when used as a system directory. This fix ensures that obvious directories are skipped.


aireilly:
LGTM

PR #231, opened 2022-06-14
Do not suggest 'discuss' for 'address' used as a noun

jhradilek:
While it is difficult to reliably recognize a noun when defining vale rules, I run a quick search on Red Hat documentatiton and it looks like a few combinations are incredibely far more common than others. Consequently, I adapted the definition to avoid flagging these word combinations.


aireilly:
LGTM

PR #233, opened 2022-06-22
Do not report recommended terms in Slash

jhradilek:
The IBM Style Guide explicitly allows N/A in situations where spelling
it out is impractical. It also explicitly recommends the use or
client/server and input/output with the slash so I added all of these to
the exceptions.

I realize that this style has ignorecase enabled but for some reason,
with vale 2.18.0, it still flagged Input/Output which is a very common
way of writing it in documentation. That is why I went ahead and used a
pattern match.


aireilly:
LGTM

PR #234, opened 2022-06-22
Do not suggest expansion of permitted acronyms

jhradilek:
These are the top acronyms that no longer require or use their original expanded form that are used in the RHEL documentation excessively.


aireilly:
LGTM

PR #269, opened 2022-08-29
fix: Make vale.ini config sample match install doc

michelle-purcell:
The sample `.vale.ini `that is provided in the [Red Hat package and repo does](https://github.com/redhat-documentation/vale-at-red-hat/blob/main/.vale.ini) not mirror the recommended configuration of the following topic:

[Installing Vale with the RedHat package](https://redhat-documentation.github.io/vale-at-red-hat/docs/user-guide/installing-vale-cli/)

image

### DIFF
--- a/.vale.ini
+++ b/.vale.ini
@@ -6,9 +6,8 @@
 # The relative path to the folder containing linting rules (styles).
 StylesPath = .vale/styles
 
-# Vocab define the exceptions to use in *all* `BasedOnStyles`.
-# See: https://docs.errata.ai/vale/vocab
-; Vocab = RedHat
+# Optional: Use the `Vocab` parameter to define exceptions to the ruleset specified for `BasedOnStyles`.
+# For more information about how to use the `Vocab` feature of vale, see: https://docs.errata.ai/vale/vocab
 
 # Minimum alert level
 # -------------------
--- a/modules/user-guide/pages/installing-vale-cli.adoc
+++ b/modules/user-guide/pages/installing-vale-cli.adoc
@@ -29,7 +29,6 @@ On the Vale site, click the tab for your operating system.
 StylesPath = styles
 
 MinAlertLevel = suggestion
-Vocab = Base
 
 Packages = RedHat
 


.vale.ini
```diff
@@ -8,7 +8,7 @@ StylesPath = .vale/styles
 
 # Vocab define the exceptions to use in *all* `BasedOnStyles`.
 # See: https://docs.errata.ai/vale/vocab
-; Vocab = RedHat
+Vocab = Base
```
michelle-purcell, line 11: The [Installing Vale with the RedHat package](https://redhat-documentation.github.io/vale-at-red-hat/docs/user-guide/installing-vale-cli/) doc recommends this configuration, but the actual file sample has a Vocab disabled. Which is the correct configuration for enabling vale-at-red-hat?

themr0c: Let's standardize on `Vocab = Base`.

The Vocabulary is a very powerful configuration that overrides the style rules.

In this repository, as we deliver a style, we must keep it empty. 

In a project, you may need to define a Vocabulary to override rules that are in your styles, without editing the styles themselves. For instance, you can continue using the Red Hat package, but modify its entire behavior. It is good for quick results, but most of the time it's better to enhance the style.

https://vale.sh/docs/topics/vocab/

themr0c:
```
E100 [vocab] Runtime error

'Base' does not exist
```

As we are here distributing a style, we don't have a Vocabulary in the repository.

To enable the Vocabulary, the required files need to be present. You need to create a `.vale/Vocab/Base` directory. To commit something in git, add empty `accept.txt` and `reject.txt` files.

I don't see this change as particularly relevant in this repository.

modules/user-guide/pages/installing-vale-cli.adoc
```diff
@@ -26,7 +26,7 @@ On the Vale site, click the tab for your operating system.
 .The `$HOME/.vale.ini` file
 [source,ini]
 ----
-StylesPath = styles
+StylesPath = .vale/styles
```
themr0c, line 29: I would suggest being consistent with https://vale.sh/hub/redhat/, which is consistent with other configurations in the Hub. Therefore, keep:

```suggestion
StylesPath = styles
```

.vale.ini
themr0c: Packages in https://vale.sh/hub/ are defined without Vocab.
The generator at https://vale.sh/generator/ is adding a Vocab line.

The Vocab being optional, I believe it's better to omit the Vocab line in our docs. Therefore remove it from https://github.com/redhat-documentation/vale-at-red-hat/blob/main/modules/user-guide/pages/installing-vale-cli.adoc.

Adding a Vocabulary is a topic for advanced users, and I believe in that case they should look at the upstream doc: https://vale.sh/docs/topics/vocab/.

modules/user-guide/pages/installing-vale-cli.adoc
themr0c: * `StylePath = styles` is consistent with https://vale.sh/hub/ and https://vale.sh/generator/. I believe we should use it as default.

* `StylePath = .vale/styles` denotes a configuration following previous recommendations in the Vale project documentation. See https://vale.sh/docs/topics/config/

* `StylePath = .github/styles` denotes a configuration following what Vale GitHub action documentation requires. See https://github.com/errata-ai/vale-action

michelle-purcell: @themr0c - Reverted 'StylesPath' back to the default configuration as per suggestion. Thanks for clarifying!

.vale.ini
michelle-purcell: @themr0c - Fixed. Vocab line is now removed from the doc sample .vale.ini configuration in this PR.

michelle-purcell:
@themr0c - Thanks for the information on Vocab and StylesPath. I have updated the diff accordingly.

themr0c (approved):

PR #275, opened 2022-08-31
Match only if YAML is used as a word

gaurav-nelson:
Currently shows error `RedHat.CaseSensitiveTerms	Use 'YAML' rather than 'yaml'.` for the line 

.Example plugin.yaml

This check ensures that the check will only match `Yaml` and `yaml` when used as words.

### DIFF
--- a/.vale/fixtures/RedHat/CaseSensitiveTerms/testinvalid.adoc
+++ b/.vale/fixtures/RedHat/CaseSensitiveTerms/testinvalid.adoc
@@ -220,3 +220,5 @@ WindowMaker
 Xemacs
 Xterm
 yaml
+Yaml is not a markup language.
+Compared to yaml, JSON has way more curly braces.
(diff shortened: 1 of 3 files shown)


jhradilek:
Hi @gaurav-nelson, I only looked briefly on your suggested change but wouldn't this mean that Vale will stop showing errors for `yaml` when it appears at the beginning of a line or when it is followed by a punctuation mark? I would imagine the following are common occurrences:

```
Yaml is not a markup language.
Compared to yaml, JSON has way more curly braces.
```

gaurav-nelson:
Thank you @jhradilek I agree with your comment. Regexes are fun 😃 

How about `\s([Yy]aml)|^([Yy]aml)` PS: https://regex101.com/r/eTohSK/1 

Let me know and I can update the PR.

themr0c:
Can you add some tests?

gaurav-nelson:
Thank you @themr0c I've updated the PR to include tests.

PR #276, opened 2022-08-31
Added a new rule to check for repeated words rule

gaurav-nelson:
Based on https://vale.sh/explorer/repeatedwords/

Matches instances of repeated words:
- This is is a sentence with a repeated word.


aireilly:
LGTM

PR #289, opened 2022-09-08
Update the ConfigMap rule and case sensitive terms for ConfigMap

gaurav-nelson:
Fixes https://github.com/redhat-documentation/vale-at-red-hat/issues/301

- Updated the regex to ignore `config` if it is part of a compound word with a dash.
- included `configmap` in the Case sensitive words.

### DIFF
--- a/.vale/fixtures/RedHat/ConfigMap/testinvalid.adoc
+++ b/.vale/fixtures/RedHat/ConfigMap/testinvalid.adoc
@@ -8,4 +8,4 @@ invalid content: config-map
 invalid content: config mail
 invalid content: config mapping
 invalid content: config abc
-invalid content: config
+invalid content: config map
--- a/.vale/fixtures/RedHat/ConfigMap/testvalid.adoc
+++ b/.vale/fixtures/RedHat/ConfigMap/testvalid.adoc
@@ -3,4 +3,4 @@ valid content: config maps
 valid content: configure
 valid content: kubeconfig
 It uses the install-config file.
-# Using config map in heading is okay
+# Using config in heading is okay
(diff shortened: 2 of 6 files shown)


.vale/fixtures/RedHat/ConfigMap/testinvalid.adoc
```diff
@@ -8,4 +8,4 @@ invalid content: config-map
 invalid content: config mail
 invalid content: config mapping
 invalid content: config abc
-invalid content: config
```
themr0c, line 11: config not followed by map is invalid content.

themr0c (changes requested):

.vale/fixtures/RedHat/ConfigMap/testinvalid.adoc
rolfedh: The Supplementary Style Guide [entry for config map](https://redhat-documentation.github.io/supplementary-style-guide/#config-map) states that the following forms are valid:
- `[cC]onfig map`
- `ConfigMap` as an API object between backticks except in headings.
-- Can we scope this variation as valid for headings only and tell users to use backticks when it appears elsewhere?

.vale/fixtures/RedHat/ConfigMap/testvalid.adoc
```diff
@@ -2,4 +2,5 @@ valid content: config map
 valid content: config maps
 valid content: configure
 valid content: kubeconfig
-
+It uses the install-config file.
```
rolfedh, line 5: For body text, this form is invalid, it must be enclosed in backticks. Can we limit this to headings only?

rolfedh (commented):
Basically, it seems like we must find a way to address two different cases: 
- one for body text where backtics are required for object names and filenames  
- and another for headings where backticks aren't allowed. 

How might we do this without allowing invalid uses in body text?

gaurav-nelson:
Closing this in favor of https://github.com/redhat-documentation/vale-at-red-hat/pull/319

PR #298, opened 2022-09-12
feat: underscores in user-defined values

themr0c:
Implements a subset of https://redhat-documentation.github.io/supplementary-style-guide/#user-replaced-values. 

The rule uses the characteristics of user-replaced values (surrounded by brackets) to search for undesired hyphens.

The pull request also adds a missing documentation page for the `RepeatedWords` rule.

I chose the suggestion level as a starting alert level. It could be useful later to increase that level if the rule proves not to create false positives.


themr0c:
@michelle-purcell This is a good example to see all required changes to add a new rule.

Issue #272, opened 2022-08-30
When "config" is preceded by "Ignition" it should be allowed


mjpytlak:
**Describe the bug**
"Ignition config" is a valid RHCOS term, but is being flagged.

Examples:

* The location to fetch the bootstrap Ignition config file from.
* Specify the infrastructure name that you extracted from the Ignition config file metadata.

**To Reproduce**

1. Add this word or expression: Ignition config
2. Vale reports an illegitimate alert: error

**Expected behavior**
"Ignition config" is not reported as an error.

**Additional context**
While the Red Hat Supplementary style guide does not list this combination as valid, I have confirmed with Kathryn Alexander that this is an approved item.

aireilly:
could we use a `Vocab` with `accept.txt` and an `[Ii]gnition config` entry to fix this?

https://vale.sh/docs/topics/vocab/

rolfedh:
@aireilly resolved, or more to do for this issue?

aireilly:
Not resolved - we don't want to use vocabs in our rules. I think we need to throw out the config map rule and create simple spelling exceptions for the various config terms we want to allow.

aireilly:
Fixed in https://github.com/redhat-documentation/vale-at-red-hat/pull/319

PR #308, opened 2022-09-14
Adding "ignition config" to spelling list

aireilly:


aireilly:
Addresses https://github.com/redhat-documentation/vale-at-red-hat/issues/272

aireilly:
This doesn't work, needs to be incorporated into the ConfigMap change I think.

PR #310, opened 2022-09-14
Adding Vocab to handle "Ignition config" term

aireilly:
Adding Vocab to handle "Ignition config" term.

Fixes https://github.com/redhat-documentation/vale-at-red-hat/issues/272


aireilly:
Closing. This is not the correct way to handle this.

PR #312, opened 2022-09-16
Recommend to avoid self-referential text

jhradilek:
In the chapter about audience and medium, the IBM Style Guides recommends to avoid self-referential text like "this chapter" or "this topic" as much as possible. As this is particularly problematic in modular documentation where such text can theoretically appear anywhere in the document hierarchy, I added a vale style to report these as a suggestion.


themr0c (approved):

aireilly:
I personally try to avoid these constructions, I like this rule! 
 
What about sentences that begin with "This procedure..." constructions? Could they be added to your list @jhradilek?

PR #319, opened 2022-09-26
Simplify rules for "config"

aireilly:
This PR removes the config map rule since it can report false positives and is likely difficult to maintain for the future as new usages of "config" will undoubtedly be allowed. Currently the SSG has "build config" and "config map" as allowed terms. "Ignition config" and "machine config" are also allowed terms.


.vale/fixtures/RedHat/Spelling/testvalid.adoc
```diff
@@ -128,6 +132,7 @@ I/O
 IDE
 IDEs
 Infinispan
+Ignition config
```
rolfedh, line 135: Because "Ignition config" is always capitalized this way, perhaps we should also test for proper capitalization.

.vale/styles/RedHat/ConfigMap.yml
```diff
@@ -1,8 +0,0 @@
----
-extends: existence
-ignorecase: true
-level: suggestion
-link: https://redhat-documentation.github.io/vale-at-red-hat/docs/reference-guide/configmap/
-message: "Do not use 'config', unless it is followed by 'map'."
```
rolfedh, line 6: This message is obsolete. 
Do we still need to provide users with some guidance to replace configmap and configuration map based on https://redhat-documentation.github.io/supplementary-style-guide/#config-map ?
Or, will the new user experience provide enough information?

rolfedh (approved):

.vale/fixtures/RedHat/Spelling/testvalid.adoc
aireilly: Added an entry to the `CaseSensitiveTerms.yml` list

rolfedh:
@themr0c any thoughts?

aireilly:
> This looks good to me. Do you want someone else to have a look and verify the changes before I merge them?

Let's wait for @gaurav-nelson or @themr0c to chime in.

themr0c (approved):

PR #320, opened 2022-09-28
Demoting crash from error to sugg

aireilly:
We currently list crash as an error term. This is incorrect. Crash in the verb form is incorrect, however, crash as a noun is acceptable in various usages. 

https://redhat-documentation.github.io/supplementary-style-guide/#crash

Demoting to Suggestion should fix for now. Ideally, we should catch with POS tagging only when crash is used as a verb. POS checking not possible with Vale rules - yet: https://github.com/errata-ai/vale-vscode/pull/60


rolfedh (approved):
Looks good to me. Ready to merge?

aireilly:
Yes please @rolfedh

PR #322, opened 2022-09-30
Add a test invalid entry for a repeated word at the end of a sentence.

### LINKED ISSUE #321: Repetition rule gets thrown off at the end of a sentence?
**Describe the bug**

Apparently, https://github.com/redhat-documentation/vale-at-red-hat/blob/main/.vale/styles/RedHat/RepeatedWords.yml gets thrown off at the end of a sentence.

Examples:

* This is a test test. <--- doesn't get flagged, because of the trailing period.

**To Reproduce**

1. Just test the sentence above.
2. Doesn't get flagged.

**Expected behavior**
Should be flagged.

**Additional context**
I haven't tested this with your full configuration- we use a different `.vale.ini`, so perhaps this doesn't happen with your full config (I fiddled a little bit copying token ignores, etc. from your `.vale.ini` and it was still a problem.

rolfedh:
Fixes #321

### DIFF
--- a/.vale/styles/RedHat/RepeatedWords.yml
+++ b/.vale/styles/RedHat/RepeatedWords.yml
@@ -1,6 +1,8 @@
 extends: repetition
 message: "'%s' is repeated!"
-level: error
+level: warning
+ignorecase: false
 alpha: true
 tokens:
+  - '[^\s\.]+'
   - '[^\s]+'
--- a/.vale/fixtures/RedHat/RepeatedWords/testinvalid.adoc
+++ b/.vale/fixtures/RedHat/RepeatedWords/testinvalid.adoc
@@ -1,1 +1,15 @@
-This is is a sentence with a repeated word.
+1. This is is a sentence with a repeated word.
+//      ----- is detected.
+
+2. This is a test test.
+//           ---------
+// `tokens` = '[^\s]+' does not detect it.
+// `tokens` = '[^\s\.]+' detects it.
+
+// 3. This is a test test,
+//              ---------
+// `tokens` = '[^\s\,]+' detects it, but neutralizes '[^\s\.]+' if placed above.
+
+// 4. Select the THE checkbox.
+//           ------- is detected if `ignorecase` is `true`.
+// However, I believe we should set it to `false`.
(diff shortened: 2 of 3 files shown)


themr0c (commented):
Why should we accept repetition in general?

.vale/styles/RedHat/RepeatedWords.yml
```diff
@@ -3,4 +3,4 @@ message: "'%s' is repeated!"
 level: error
 alpha: true
 tokens:
-  - '[^\s]+'
+  - '(\b\S+\b)\s+\b\1\b'
```
rohennes, line 6: Vale uses golang flavour of regex which doesn't support lookbacks (the "\1" part). Try this: `[^\s\.]+`
It seems to work, although I can't fully understand the regex tbh

alexpdp7: That works perfectly with our Vale configuration, thanks!

rolfedh:
> Why should we accept repetition in general?

I don't think the `RedHat` style should accept repetition. I think it should catch it as an error, or at least as a warning. This update is intended to fix the false negative that @alexpdp7 discovered and raised in https://github.com/redhat-documentation/vale-at-red-hat/issues/321

.vale/styles/RedHat/RepeatedWords.yml
rolfedh: > Try this: `[^\s\.]+`

Thank you, @rohennes !

rolfedh:
I'd appreciate your reviews, @alexpdp7, @rohennes, and @themr0c. Thanks!

alexpdp7 (approved):
This passes the following test cases (using our own fixture mechanism):

```
Select the the checkbox.
       ------- 'the' is repeated!

# In enumerations, commas should "break up" repetition.
Allow searches to be performed on type, contributor, contributor type, and others.
                                        ^^^^^^^^^^^^^^^^^^^^^^^^

# some variations of this rule didn't work correctly with trailing punctuation.
This is text text.
        --------- 'text' is repeated!

# This is a false positive, but capitalization needs to match, so I think it's a non-issue
This is a Text. Text is nice.
          ---------- 'Text' is repeated!

# The "normal" case works
This is a text. Text is nice.
          ^^^^^^^^^^
```

Meaning the rule fails on the `---` bits, and doesn't complain on the `^^^`. It has a false positive, but that should be rare enough not to be a concern. So I would recommend merging in this change.

Issue #326, opened 2022-10-03
Vale rule about `ConfigMap` is an error


max-cx:
**Describe the bug**

The word `ConfigMap` triggers an illegitimate Vale error:
![image](https://user-images.githubusercontent.com/78564304/193532218-01381731-266f-46fb-83ee-ae7eb7644cec.png)
The supplementary style guide does not tell us to use `config map` rather than `ConfigMap`, so the error and its message are incorrect.

I've looked up the [supplementary style guide](https://redhat-documentation.github.io/supplementary-style-guide/#_c), and
a) `ConfigMap` is stated in the description for `config map`: `The API object for a config map is ConfigMap.`.
b) `ConfigMap` is not listed among the incorrect forms for `config map`: `Incorrect forms: configmap, configuration map`

![image](https://user-images.githubusercontent.com/78564304/193530938-91c107c7-d4ff-4081-a3c4-b12e1212fb46.png)

**To Reproduce**
1. Add `ConfigMap` in a text.
2. Vale reports an illegitimate alert: error.

**Expected behavior**
1. Make the trigger word in the rule to be case-sensitive: only the lower-case `configmap`, and not `ConfigMap`.
2. Retain the error level of the alert.
3. The error message has to offer the Vale user both alternatives, something like this:
```
Use either 'config map' or '`ConfigMap`' as appropriate to your text.
```
4. Add a new Vale rule for `ConfigMap`: only for the Pascal case without back ticks, and not `configmap`.
5. Make the new rule for `ConfigMap` in the previous point an error.
6. The error message has to ask the Vale user to verify that they really meant the Pascal-case `ConfigMap`, something like this:
```
If 'ConfigMap' refers to an API object, enclose it in back ticks: '`ConfigMap`'.
```
By the way, what happens to occurrences of 'ConfigMap' without back ticks inside 'yaml' code blocks? Will those be treated as errors by Vale?

**Additional context**
Please prioritize this issue because it's blocking a PR with multiple instances of this error.

aireilly:
I think this issue can be marked as invalid/won't fix - `ConfigMap` as an API reference should be marked up with backticks, which would remove the error. Open to correction on this.

max-cx:
> ConfigMap as an API reference should be marked up with backticks

The current error message instructs the Vale user to write it as 'config map', contrary to your advice.
If this is a won't fix, then I'll have to follow the Vale instruction and fix all errors as 'config map', not '`ConfigMap`'.

[...]

PR #327, opened 2022-10-03
Don't fail on ConfigMap

aireilly:
Addresses https://github.com/redhat-documentation/vale-at-red-hat/issues/326


aireilly:
Won't pass until https://github.com/redhat-documentation/vale-at-red-hat/pull/330 gets merged.

aireilly:
I've totally reworked this, dropping all the Pascal case changes. This is more straightforward.

PR #329, opened 2022-10-03
Simplify YAML regex

aireilly:
This PR simplifies the regex to catch `Yaml`, `yaml`, but not `xyz.yaml`.


aireilly:
Won't pass until https://github.com/redhat-documentation/vale-at-red-hat/pull/330 gets merged.

PR #334, opened 2022-10-04
Pascal and Camel case rule

aireilly:
Pascal and Camel case rule that suggests marking up matching terms in body text in backticks. Does not trigger in headings or codeblocks, etc.


.vale/styles/RedHat/PascalCamelCase.yml
```diff
@@ -0,0 +1,51 @@
...
+  - Build
+  - Config
+  - Console
+  - Deployment
+  - Endpoints
+  - Etcd
+  - Event
+  - Group
+  - Help
+  - Identity
```
jhradilek, line 25: I am not sure how I feel about common words that are often used with upper case in project names. This, for example, would trigger a lot of false positives for Identity Management (IdM) in RHEL documentation (422 false positives to be precise).

.vale/styles/RedHat/PascalCamelCase.yml
```diff
@@ -0,0 +1,51 @@
...
+  - Image
```
jhradilek, line 26: This would trigger a lot of false positives for Image Builder in RHEL documentation (484).

.vale/styles/RedHat/PascalCamelCase.yml
```diff
@@ -0,0 +1,51 @@
...
+  - Infrastructure
```
jhradilek, line 27: This would trigger false positives for Red Hat Update Infrastructure (RHUI) in RHEL documentation.

.vale/styles/RedHat/PascalCamelCase.yml
```diff
@@ -0,0 +1,51 @@
...
+  - Project
+  - Prometheus
+  - Provisioning
+  - Proxy
+  - Role
+  - Route
+  - Scheduler
+  - Secret
+  - Storage
+  - Subscription
```
jhradilek, line 46: This would trigger a lot of false positives for Subscription Manager in RHEL documentation (78).

.vale/styles/RedHat/PascalCamelCase.yml
```diff
@@ -0,0 +1,51 @@
...
+  - Profile
+  - Project
+  - Prometheus
+  - Provisioning
+  - Proxy
+  - Role
+  - Route
+  - Scheduler
+  - Secret
+  - Storage
```
jhradilek, line 45: This would trigger false positives for any mentions of Red Hat Ceph Storage, Red Hat Gluster Storage, Red Hat Enterprise Linux Resilient Storage, and similar.

.vale/styles/RedHat/PascalCamelCase.yml
```diff
@@ -0,0 +1,51 @@
...
+  - Group
+  - Help
+  - Identity
+  - Image
+  - Infrastructure
+  - Ingress
+  - Lease
+  - Namespace
+  - Network
+  - Node
```
jhradilek, line 32: I haven't tested it but wouldn't this also trigger for all occurrences of `Node.js`?

.vale/styles/RedHat/PascalCamelCase.yml
aireilly: @jhradilek Thanks for digging in! I figured this would be tricky for these single word terms, which is why I am hesitant to merge the PR as is.

.vale/styles/RedHat/PascalCamelCase.yml
aireilly: Yes it would :/

.vale/styles/RedHat/PascalCamelCase.yml
aireilly: Great feedback - I'll rethink this one ;)

aireilly:
@themr0c WDYT of this PR?

themr0c:
It's a new rule? Is it missing a page in the reference guide? The local test script should have created it.

rohennes:
If this PR gets refined, then #355 might be another one to add to the list of terms

aireilly:
I've made updates based on @jhradilek's and @rohennes' suggestions. I think this is ready to merge. @rolfedh @themr0c WDYT? The only possible controversial items that will be flagged for wrapping in backticks are this case sensitive list:

  - Config
  - ConfigMap
  - Deployment
  - Event
  - Group
  - Help
  - Image
  - Ingress
  - Lease
  - Namespace
  - Network
  - Node(?!\.js)
  - OAuth
  - Pod
  - Probe
  - Profile
  - Project
  - Prometheus
  - Provisioning
  - Proxy
  - Role
  - Route
  - Scheduler
  - Secret
  - Storage
  - Template
  - Tuned

.vale/fixtures/RedHat/PascalCamelCase/testinvalid.adoc
```diff
@@ -0,0 +1,148 @@
...
+DaemonSet
+Deployment
+DeploymentConfig
+DNSRecord
+EgressNetworkPolicy
+EndpointSlice
+Event
+FeatureGate
+FlowSchema
+Group
```
rohennes, line 45: vSphere Port Group
Master Security Group ID
Linux Control Group

Some false positives I believe

.vale/fixtures/RedHat/PascalCamelCase/testinvalid.adoc
```diff
@@ -0,0 +1,148 @@
...
+MachineConfig
+MachineConfigPool
+MachineHealthCheck
+MachineSet
+MutatingWebhookConfiguration
+NetNamespace
+NetworkAttachmentDefinition
+NetworkPolicy
+newCustomerId
+Node
```
rohennes, line 83: Node Feature Discovery
Single Node OpenShift

Is the word Node going to trigger this rule? I'd leave it out if so

rolfedh:
@aireilly, from your [comment here](https://coreos.slack.com/archives/C0218RXJK5E/p1664979310847379?thread_ts=1664971025.830069&cid=C0218RXJK5E), we might to put this new rule through a "training process" where we test it against a large body of files and add any false positives to the exceptions block. WDYT?

rolfedh:
I think @michaelryanpeter and others tested VaRH in the openshift-docs repo and gathered the errors in a spreadsheet.
Perhaps we could do something similar: demote this rule to a suggestion, gather data from the output of one or more repos, review the data, and update the exception block. When we're satisfied with the quality of the output, we could promote it from being a suggestion to a warning.

michaelryanpeter:
I believe that @gaurav-nelson and @aireilly were behind the idea and implementation of the rule audit and associated spreadsheet.

michaelryanpeter:
FWIW, I second the idea for doing an audit of the rule. On the one hand, it would be a great improvement to the OpenShift docs to catch the terms that are used as proper nouns in Kubernetes. On the other hand, the false positives issue is a concern. As a caveat, I have not had a chance to test this change or look closely at the rules.

aireilly:
Good idea to test further. 
```
$ grep 'Config \| ConfigMap \| Deployment \| Event \| Group \| Ingress \| Lease \| Node \| OAuth \| Pod \| Probe \| Prometheus \| Role \| Scheduler \| Secret \| Storage \| Template \| Tuned ' -r --include=**\*.adoc
```

should give a better idea per repo what is being caught.

aireilly:
Also could just drop these single word terms from this rule altogether. 🤷

.vale/fixtures/RedHat/PascalCamelCase/testinvalid.adoc
```diff
@@ -0,0 +1,148 @@
...
+OAuthClient
+OAuthClientAuthorization
+OpenShiftAPIServer
+OpenShiftControllerManager
+OperatorGroup
+OperatorHub
+OperatorPKI
+PackageManifest
+PersistentVolume is an API object.
+Pod
```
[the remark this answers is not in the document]
aireilly: I think I'll just remove these single words from this rule altogether.

rohennes:
> Also could just drop these single word terms from this rule altogether. shrug

I would lean towards that. This rule targeted compound words using Camel/Pascal case anyway, right? For single words it's harder to identify if they are objects/commands. 

For example, "Infra nodes are used to host Monitoring, Ingress, and Registry components..." TBH I'm always sketchy on the backticks rule, but I don't think Ingress needs backticks here and it's kinda impossible to create a rule smart enough to let this pass?

michaelryanpeter:
In the OpenShift docs, Ingress should be capitalized without backticks. See the [OCP repo glossary](https://github.com/openshift/openshift-docs/blob/30dc46d020d38222dea3277536910424be756775/contributing_to_docs/term_glossary.adoc#ingress). There isn't guidance there for _Monitoring_ or _Registry_.

.vale/styles/RedHat/PascalCamelCase.yml
```diff
@@ -0,0 +1,42 @@
+---
+extends: existence
+scope: sentence
+ignorecase: false
+nonword: true
+level: warning
```
rolfedh, line 6: Do we want to change this to `suggestion` temporarily while we put this through a testing phase?

aireilly:
Demoted rule to suggestion as suggested by @rolfedh

Issue #333, opened 2022-10-04
Broken formatting in the linting-vale-cli page


themr0c:
Broken content in https://redhat-documentation.github.io/vale-at-red-hat/docs/user-guide/linting-vale-cli/:

* The code block is misaligned
* Erroneous `+` before "Where", and this block is misaligned
* The space after the double dashes for the option is misleading.

![Screenshot from 2022-10-04 13-25-26](https://user-images.githubusercontent.com/243761/193807640-bfc9fe74-81de-4590-ae26-4fb944109a75.png)

rolfedh:
https://redhat-documentation.github.io/vale-at-red-hat/docs/user-guide/using-vale-cli/

PR #336, opened 2022-10-04
Fix floating plus

rolfedh:


rolfedh:
Closes https://github.com/redhat-documentation/vale-at-red-hat/issues/333

Issue #208, opened 2022-04-25
'by using' not detected, and is incorrectly flagged as a RH style suggestion


michelle-purcell:
**Describe the bug**

vale-at-red-hat flags a suggestion for content that is already style guide compliant.

' by using'  triggers an illegitimate vale suggestion to use 'by using'

**To Reproduce**

1. Add this word or expression: 'by using'
2. Vale reports an illegitimate alert: error, warning, suggestion.

![image](https://user-images.githubusercontent.com/92924207/165092018-fefd7659-fc8e-4506-9e44-e9a32d5e186c.png)

**Expected behavior**
the linter should ignore this line because 'by' already precedes 'using'.

rohennes:
I wasn't able to replicate this one @michelle-purcell , is it still an issue?

michelle-purcell:
Thanks, @rohennes for testing.
I just retested this one and it's OK now. I think we can put this down to an intermittent issue with the VS Code extension as I can no longer reproduce it. I'll reopen if it reoccurs. :+1:

rohennes:
Something odd going on there. I think it's related to the environment where by using occurs. I can replicate this by shuffling the sentences around in the testvalid.adoc:
![image](https://user-images.githubusercontent.com/104497497/194531022-6ae338b3-a16c-4063-a99d-2e0c3667fa05.png)
![image](https://user-images.githubusercontent.com/104497497/194531083-ac799e86-250a-4b69-89a2-71f60df10d49.png)
Moving the content around moves the error. However if I change the style to add .* at the beginning of the regex, it fixes the issue: `".*(?<!by) using": by using|that uses`
![image](https://user-images.githubusercontent.com/104497497/194531213-b09d2e3b-6cd0-47c8-b7db-1c00bf88ac9a.png)
I'll open a PR and see if others can test it.

rohennes:
https://github.com/redhat-documentation/vale-at-red-hat/pull/362/files - still being weird so had to rollback the change I made :-(

aireilly:
Think we need a sequence rule for this one: https://vale.sh/docs/topics/styles/#sequence

rohennes:
Opened a Vale issue: https://github.com/errata-ai/vale/issues/674

PR #344, opened 2022-10-07
'By using' gets incorrectly flagged depending on the surrounding context

rohennes:
Addresses #208.

I think the term 'by using' gets flagged incorrectly depending on the surrounding context. If you play around with the testvalid.adoc for termsuggestions, you can generate an incorrect error on 'by using'. 

image

See issue for more examples. I think it relates to what precedes the instance of 'by using'. Adding regex to allow anything before 'by' seems to fix this issue.


michelle-purcell:
@rohennes - Great investigative work here Ronan!  Thank you :+1:   LGTM

aireilly:
LGTM

Issue #293, opened 2022-09-12
Replace links to https://vale.sh/docs


themr0c:
Links to the Vale documentation should point to the new website: https://vale.sh/docs, rather than pointing to the obsolete errata.ai website.

Additionally, some documents have moved. Verify the linked URLs provides the desired content.

PR #345, opened 2022-10-07
Updated URLs to https://vale.sh/docs/

rolfedh:
Closes https://github.com/redhat-documentation/vale-at-red-hat/issues/293


rolfedh:
Ran "linkie" link checker to verify that none of the URLs in our docs are broken.

PR #351, opened 2022-10-10
Jetbrains -> JetBrains

rkratky:
typo fix


aireilly:
Can you update the text fixtures as well please.

rolfedh:
@aireilly @rkratky what do you think about the proposed changes in https://github.com/redhat-documentation/vale-at-red-hat/pull/361 instead.

aireilly:
Addressed in https://github.com/redhat-documentation/vale-at-red-hat/pull/361

PR #362, opened 2022-10-19
Rolling back changes to the 'by using' term

rohennes:
#344 introduced an error where the whole sentence got flagged.


aireilly:
LGTM

Issue #356, opened 2022-10-12
"which" is sometimes incorrectly flagged by RedHat.TermsSuggestions


jafiala:
**Describe the bug**
"which" can be used correctly without a comma when preceded by a preposition.

Examples:

* Install the `shmackage` package on the systems **on which** you want to install the server.

**To Reproduce**
Enter the example sentence.

**Expected behavior**
I would expect "which" to not be flagged when preceded by a preposition (on, to, for, in, with, against, without, ...?)

PR #366, opened 2022-10-20
Fix issue #356: Which is sometimes incorrectly flagged

jhradilek:
Fixes issue #356.

### DIFF
--- a/.vale/fixtures/RedHat/TermsSuggestions/testvalid.adoc
+++ b/.vale/fixtures/RedHat/TermsSuggestions/testvalid.adoc
@@ -13,3 +13,13 @@ such as
 write
 this clause that is restrictive
 this clause, which is nonrestrictive
+something in which somehing else
+something on which something else
+something at which something else
+something to which something else
+something for which something else
+something from which something else
+something of which something else
+something with which something else
+something without which something else
+something against which something else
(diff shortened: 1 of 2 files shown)


aireilly (commented):
LGTM, this is a great addition to the rules.

aireilly:
One small issue: in Your PR, if you add a line which _does_ have an error above your valid "which" examples, Vale highlights an incorrect line as having an error. This is fixed by adding `scope: sentence` to `TermsSuggestions.yml`. I am not sure whether this is a bug or limitation in Vale tbh. WDYT @jhradilek?

image

jhradilek:
Hi @aireilly, this is a very good observation, I should have caught it: I briefly noticed some odd behavior while fine tuning the regular expression (missing lines in the `testinvalid.adoc`) but failed to recognize it. I will look closer in the evening, but I am not sure why this particular regular expression would behave differently so I wonder if this problem exists for other styles that extend substitution like `SimpleWords.yml`.

aireilly:
I think adding `scope: sentence` to this PR would allow us to merge. WDYT @jhradilek?

aireilly:
Ah I think the issue above is caused by how Asiidoc is rendered into HTML for Vale to parse. AsciiDoc uses carriage return + blank line for paragraphs. Vale runs on the generated HTML. Carriage returns on their own do not equate to sentences. So the regex sees one long sentence in the HTML output. Background: https://writetheasciidocs.netlify.app/ventilated-prose

aireilly:
Closing this as addressed in https://github.com/redhat-documentation/vale-at-red-hat/pull/366

aireilly:
See https://github.com/errata-ai/vale/issues/674

Issue #365, opened 2022-10-20
Add "leverage" to the use with caution list


aireilly:
https://www.ibm.com/docs/en/ibm-style?topic=word-usage

Use with caution: **leverage**   (verb)

Do not use in technical information because it does not translate correctly in some languages.

In marketing content, do not use as a synonym for "use", "exploit", or "take advantage of”. Use only in the context of using something in a way that multiplies its effect as by a lever, such as using borrowed capital to multiply the return on an investment.

kquinn1204:
@aireilly The term `leverage` is already defined in `TermsErrors.yml` with substitute value set to `use`. Will I remove from there and TermsErrors/testinvalid.adoc and instead add to `Usage.yml` and `Usage/testinvalid.adoc`

aireilly:
Hmm. I'm second guessing myself now. I agree that this can stay as an error, particularly in the context of our content. No changes required. Closing.

PR #367, opened 2022-10-21
adding leverage to usage

kquinn1204:


rolfedh:
Hi, Kevin. 

What kind of additions do you have in mind?

I see that "leverage" is covered by line 184 in `.vale/styles/RedHat/TermsErrors.yml`:
```
  leverage: use
```

This rule is based on an entry in the _IBM Style Guide_ In  2020 that was last updated in January 2022:
```
leverage verb
Do not use in technical information because it does not translate correctly in some languages.
In marketing content, do not use as a synonym for "use", "exploit", or "take advantage of”. Use only in the
context of using something in a way that multiplies its effect as by a lever, such as using borrowed capital to
multiply the return on an investment.
```

aireilly:
https://github.com/redhat-documentation/vale-at-red-hat/issues/365 I raised this, thinking that leverage could move to a warning, but now I'm not sure. I think it is OK as an error

.vale/styles/RedHat/TermsSuggestions.yml
```diff
@@ -25,6 +25,7 @@ swap:
   crash: fail|lock up|stop|stop responding
   functionality: functions # IBM
   input|type: enter (followed by the text to enter in monospace) # https://redhat-documentation.github.io/supplementary-style-guide/#text-entry
+  may: might
```
rolfedh, line 28:
```suggestion
  may: might (for possiblity)|can (for ability) 
```

rolfedh:
```
might verb
Use "might" to indicate possibility. To indicate ability, use "can".
In most cases, use "can" or "might", not "may".
```

rolfedh (approved):
Good addition, Kevin. 
I thought we might expand the suggestions to include "can". WDYT?

rolfedh:
Do you think we should move this to TermsWarnings?

PR #381, opened 2022-11-10
chore: add YAML delimiter in all rules

themr0c:
Add YAML delimiter (`---` line).

This is useful to concatenate all rules into a single YAML file for use with Grazie Professional on IntelliJ editors. See: https://plugins.jetbrains.com/plugin/16136-grazie-professional/docs/project-style-guides.html


aireilly:
LGMT

PR #383, opened 2022-11-11
Remove plugin from disallowed term list

michelle-purcell:
The Red Hat Style Council recently revisited and updated the guidelines for using `plugin` and `plug-in` (noun & adjective forms) in Red Hat documentation, deviating from the ISG.

As per [RHSS issue 213](https://github.com/redhat-documentation/supplementary-style-guide/issues/213), `plugin` is  now the preferred form:

`Use "plugin" rather than "plug-in", unless you are updating existing content that uses the hyphenated form.`

This PR removes `plugin` from the Terms Errors rule and adds `plug-in` to the Terms Warnings rule.

This PR also removes the plural form, which was previously added to some of the rules/fixtures.

### DIFF
--- a/.vale/styles/RedHat/TermsErrors.yml
+++ b/.vale/styles/RedHat/TermsErrors.yml
@@ -219,7 +219,7 @@ swap:
   perfcounter: performance counter
   perimeter network: DMZ
   plaintext|plain-text|cleartext|clear text: plain text
-  plugin|plugins|plug in|plug ins: plug-in # https://www.merriam-webster.com/dictionary/plug-in
+  plug in|plug ins: plugin
   power down: turn off
   power off: turn off
   power on: turn on
--- a/.vale/fixtures/RedHat/TermsErrors/testinvalid.adoc
+++ b/.vale/fixtures/RedHat/TermsErrors/testinvalid.adoc
@@ -256,8 +256,6 @@ perimeter network
 phone
 plain-text
 plaintext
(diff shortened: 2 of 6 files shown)


.vale/styles/RedHat/TermsErrors.yml
```diff
@@ -219,7 +219,7 @@ swap:
   perfcounter: performance counter
   perimeter network: DMZ
   plaintext|plain-text|cleartext|clear text: plain text
-  plugin|plugins|plug in|plug ins: plug-in # https://www.merriam-webster.com/dictionary/plug-in
+  plug in|plug ins: plugin # https://www.merriam-webster.com/dictionary/plug-in
```
aireilly, line 222: Nit: Can we remove this comment? We don't need it here and it's confusing.

aireilly (commented):
One small nit, otherwise LGTM :)

rolfedh (approved):
When you remove the comment, this will be ready to merge.

.vale/styles/RedHat/TermsErrors.yml
michelle-purcell: @aireilly - Good catch. Thanks. Funny how my brain has been trained to totally ignore stuff that's remmed out!
Noise removed :-)

aireilly: Merged :)

PR #386, opened 2022-11-16
Do not report common image formats

jhradilek:
This is a minor thing but while doing a peer review, I noticed that Vale complains about common image formats like `GIF`, `PNG`, and `JPEG`. I contemplated adding the `JPG` variant too but as it is technically not an acronym and was used as a file extension only because of early file system limitations, I chose not to.


aireilly:
LGTM

PR #394, opened 2022-12-02
Demote verb error on JAR to a warning

michelle-purcell:
JAR is incorrectly flagged as an error against non-verb word usage in the vale report with the guidance:  "Use 'compress' or 'archive' rather than 'JAR'. 

image

In JAVA-related product docs, a task with lots of steps that include working with a JAR file, creates a lot of false-positive errors.

The style guideline for 'jar' relates to using jar/JAR as a verb, for example:
 **To _jar_ something** :

image

This PR demotes the error to a suggestion and adds (verb) clarification. The PR also adds "JAR file" as a permitted exception.

### DIFF
--- a/.vale/styles/RedHat/TermsErrors.yml
+++ b/.vale/styles/RedHat/TermsErrors.yml
@@ -169,7 +169,6 @@ swap:
   insure: ensure
   Internet address: IP address|URL|Internet email address|web address
   irrecoverable: unrecoverable
-  jar: compress|archive
   keep in mind: remember
   kernelspace: kernel-space
   labour: labor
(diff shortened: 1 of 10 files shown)


.vale/styles/RedHat/TermsErrors.yml
```diff
@@ -170,7 +170,6 @@ swap:
   insure: ensure
   Internet address: IP address|URL|Internet email address|web address
   irrecoverable: unrecoverable
-  jar: compress|archive
```
aireilly, line 173: So with this change, we're effectively allowing "jar/unjar" as a verb?

michelle-purcell:
@aireilly - All build errors are fixed. Think we are good to go :+1:

PR #404, opened 2023-01-06
Enhance passive voice rule message.

michelle-purcell:
**Problem:**
The existing message that is presented when passive voice is detected suggests only one of the acceptable uses that our style guidelines call out. 

image

This PR adjusts the detail in the message to remove the detailed prerequisites guidance:

image


michelle-purcell:
Thanks @dfitzmau :+1:

aireilly:
LGTM

PR #405, opened 2023-01-06
Update passive voice docs to be more descriptive

aireilly:


michelle-purcell (approved):
LGTM :+1: 

But I do wonder whether we need to rethink this Reference section and its purpose. 

It makes sense to list and introduce the rulesets but IMHO, we need to be careful not to provide too much detail about each specific style guideline (ssg+ibm) as it'll be a challenge to keep it up-to-date.

PR #409, opened 2023-01-17
Adding rule to flag IPI, UPI and AI acronyms

rohennes:
We should not use UPI or IPI acronyms in the docs according to the [SSG](https://redhat-documentation.github.io/supplementary-style-guide/). Instead use "installer-provisioned infrastructure" and "user-provisioned infrastructure". 

I am pretty certain the same goes for Assisted Installer although there is no official guidance. I opened an issue to follow up with that. I did check with the Assisted Installer writers and AI should be avoided - obvious confusion with artificial intelligence. I think it's pretty safe to add AI here too.

### DIFF
--- a/.vale/fixtures/RedHat/TermsErrors/testinvalid.adoc
+++ b/.vale/fixtures/RedHat/TermsErrors/testinvalid.adoc
@@ -8,6 +8,7 @@ a transition
 abort
 administrate
 advisor
+AI
 air wall
 all caps
 all right
@@ -194,6 +195,7 @@ information on
 information technology
 insure
 Internet address
+IPI
 irrecoverable
 keep in mind
 kernelspace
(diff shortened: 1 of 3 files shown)


osslate:
As far as I can remember, we want to limit VaRH rules to already defined guidelines in the SSG. It might be a good idea to pass this by the SSG team before merging the "AI" addition if it isn't present there yet.

rohennes:
> As far as I can remember, we want to limit VaRH rules to already defined guidelines in the SSG. It might be a good idea to pass this by the SSG team before merging the "AI" addition if it isn't present there yet.

Thanks Fioinn, just got it merged yesterday! :-) 
https://github.com/redhat-documentation/supplementary-style-guide/pull/276 

One issue I had in the current PR was I couldn't seem to specify upper-case only for the acronyms, so this catches lower-case instances of ipi/upi/ai too...

aireilly:
LGTM

PR #411, opened 2023-01-23
Add mTLS to Headings

rolfedh:
Fixes https://github.com/redhat-documentation/vale-at-red-hat/issues/412


michelle-purcell:
@rolfe - LGTM. Thank you for adding this.

Do we also need to add an entry into the CaseSensitiveTerms.yml too to catch instances of `MTLS` in body text, which writers might be tempted to use also? For example: 

`MTLS|m-TLS: mTLS`

michelle-purcell (approved):

PR #416, opened 2023-01-31
Add a warning for words 'sane' and 'sanity'

jhradilek:
In the Word Usage section, the IBM Style Guide recommends to avoid the use of word 'sane' as it might be offensive to neurodiverse people. As it provides alternatives, I added it to TermsWarnings to provide a warning.

In addition, the IBM Style Guide recommends to avoid using the phrases 'sanity check' and 'sanity test'. There already is a rule for 'sanity check' but for completeness and to honor the intent behind these recommendations, I added a warning for 'sanity test' and 'sanity' as well.

### DIFF
--- a/.vale/fixtures/RedHat/TermsErrors/testinvalid.adoc
+++ b/.vale/fixtures/RedHat/TermsErrors/testinvalid.adoc
@@ -285,6 +285,7 @@ run level
 run-level
 S 390
 sanity check
+sanity test
 screensaver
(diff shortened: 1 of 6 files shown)


aireilly:
LGTM. OK to merge?

jhradilek:
Hi @aireilly, in case the question is addressed to me, unless I missed something that I am unaware of, this merge request is complete from my side.

PR #445, opened 2023-02-20
Ensure the correct spelling of OpenShift

jhradilek:
An unfortunate typo sadly occasionally makes it to published documentation. Adding it to Vale increases author's chances of catching it early.


themr0c (approved):

themr0c:
LGTM
Be mindful that `TermsErrors` is a case insensitive rule.

PR #451, opened 2023-02-20
feat: 'Command Prompt' is correct, 'command prompt' is incorrect

### LINKED ISSUE #436: Command Prompt is correct, command prompt is incorrect
Command Prompt is a command name in Windows:

* https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/c/command-prompt
* https://support.microsoft.com/en-us/windows/command-prompt-and-windows-powershell-for-windows-11-6453ce98-da91-476f-8651-5c14d5777c20

Therefore, the following Vale rule require an amendment to match only in lowercase:

```
RedHat.TermsErrors:Use 'shell prompt' rather than 'Command Prompt'.
```

We should  have:

* No error for 'Command Prompt'
* Error for 'command prompt'

The style guide states in https://redhat-documentation.github.io/supplementary-style-guide/#_s:

```
shell prompt (noun)

Description: The shell prompt is the character at the beginning of the command line, for example "$" or "#". It indicates that the shell is ready to accept commands. Do not use "command prompt", "terminal", or "shell".

Use it: yes

Incorrect forms: command prompt, terminal, shell
```

themr0c:
fixes #436

### DIFF
--- a/.vale/styles/RedHat/CaseSensitiveTerms.yml
+++ b/.vale/styles/RedHat/CaseSensitiveTerms.yml
@@ -76,7 +76,7 @@ swap:
   cidr|Classless Interdomain Routing|Classless Inter-domain Routing: CIDR
   classic mode: GNOME Classic
   CloudForms Management Engine|CFME: Red Hat CloudForms|Red Hat CloudForms Appliance
-  command prompt: Command Prompt|shell prompt
+  command prompt: shell prompt|Command Prompt (Windows product)
   Composite Content view|composite content view|Composite View|composite view: Composite Content View
   configmap: config map
   configuration map: config map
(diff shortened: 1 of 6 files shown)


.vale/styles/RedHat/CaseSensitiveTerms.yml
```diff
@@ -76,6 +76,7 @@ swap:
   cidr|Classless Interdomain Routing|Classless Inter-domain Routing: CIDR
   classic mode: GNOME Classic
   CloudForms Management Engine|CFME: Red Hat CloudForms|Red Hat CloudForms Appliance
+  command prompt: Command Prompt|shell prompt
```
rolfedh, line 79: Without some context, folks might replace "command prompt" with "Command Prompt" everywhere. Would adding these parenthetical comments help?
```suggestion
  command prompt: Command Prompt (Windows)|shell prompt (Linux and elsewhere)
```

rolfedh (approved):
LGTM. Offered a suggestion.

.vale/styles/RedHat/CaseSensitiveTerms.yml
themr0c: shell prompt first, then specify that Command Prompt is the name of a Windows product?

```suggestion
  command prompt: shell prompt|Command Prompt (Windows product)
```

PR #452, opened 2023-02-21
Turn off PascalCamel case rule in headings

aireilly:
Turn off PascalCamel case rule in headings by using `scope: [list, sentence]` in the rule.


themr0c (approved):
That's a nice improvement.

aireilly:
> That's a nice improvement.

Yeah I think it could be useful in other contexts too. I have raised [this](https://github.com/errata-ai/vale/issues/561) but being able to specify an array of scope values probably meets this requirement from the other side.

PR #463, opened 2023-03-01
improve Definitions.yml

aireilly:
Definitions.yml rule should not fire when there is an explanation immediately following in parenthesis.

image


rolfedh (approved):
LGTM

Issue #462, opened 2023-02-28
vale reports an error for referring to x86_64 rather than AMD64, even when describing an x86_64 command


mnsteven:
The command to enable a repository for the various cluster and resilient storage repositories includes x86_64 (this one for rhel 8 resilient storage):

subscription-manager repos --enable=rhel-8-for-x86_64-resilientstorage-rpms

When I describe that command, I note that this is the example to use for an x86_64 system.  Vale tells me that this is an error.  But to say to use this command for an AMD64 system seems confusing there.  I checked with the technical PM on the project who said:

"that error message is not valid in this case.

Maybe if you were to refer to the x86_64 architecture outside of actual commands they want you to refer to it as AMD64, but not in the --enable command."

So you could say that when I'm introducing the command I am referring the to architecture outside of the actual command, but it still seems confusing in this case and I'm not sure it would be helpful to the user to call the architecture one thing in the text and another thing in the command/repository name.

aireilly:
```asciidoc
[source,terminal]
----
$ subscription-manager repos --enable=rhel-8-for-x86_64-resilientstorage-rpms
----
```

When we describe commands in our docs, the AsciiDoc style requires commands to be marked up as codeblocks with appropriate huighlighting applied. See: https://redhat-documentation.github.io/asciidoc-markup-conventions/

aireilly:
Having said that, there is a long standing issue on the [Supplementary Style Guide GH repo](https://github.com/redhat-documentation/supplementary-style-guide/issues/171) which I am following. I think we will need to update our rules based on the outcome of this.

mnsteven:
The problem is not in the codeblock, but in the description that precedes the codeblock:

"Enable the repository for Resilient Storage that corresponds to your system architecture. For example, to enable the Resilient Storage repository for an x86_64 system, enter the following subscription-manager command: "

As noted in my description:

"So you could say that when I'm introducing the command I am referring the to architecture outside of the actual command, but it still seems confusing in this case and I'm not sure it would be helpful to the user to call the architecture one thing in the text and another thing in the command/repository name."

aireilly:
Ah OK, my bad. I understand. I think we can adjust this and not report an error. I'll submit a PR.

aireilly:
CLosing this, demoted the x84_64 rule to a suggesiton. Hope this works.

mnsteven:
[...]

Issue #199, opened 2022-03-16
Use 'AMD64' rather than 'x86_64' (RedHat.TermsErrors)


max-cx:
**Describe the bug**

This word or expression triggers an illegitimate vale alert:
**error  Use 'AMD64' rather than         RedHat.TermsErrors 
               'x86_64'.** 

I recommend removing this rule for the following reasons:

`x86_64` and `AMD64` are not interchangeable. `x86_64` covers both `Intel 64` and `AMD64`. To rewrite `x86_64` as `AMD64` means telling readers (including those that use `Intel 64` for Red Hat products) that `Intel 64` does not exist and is not supported. (In other words, with this rule Vale tells us that `car` is an error and that we must replace `car` with `BMW`.)

OpenShift docs (and many, many package names for Linux distros) refer to `x86_64`: https://docs.openshift.com/container-platform/4.10/installing/installing-preparing.html

More info below:
https://en.wikipedia.org/wiki/List_of_Intel_CPU_microarchitectures#64-bit_(x86-64)
https://en.wikipedia.org/wiki/X86
https://www.pcmag.com/encyclopedia/term/x64
https://www.intel.com/content/www/us/en/architecture-and-technology/microarchitecture/intel-64-architecture-general.html
https://www.pcmag.com/encyclopedia/term/amd64 (edited) 

![image](https://user-images.githubusercontent.com/78564304/158636550-13da6904-40a1-4173-b81c-0d7eda79ebdf.png)
(source: https://www.intel.com/content/www/us/en/architecture-and-technology/microarchitecture/intel-64-architecture-general.html)

`Intel 64` is explicitly supported on RHEL:
https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/performing_a_standard_rhel_installation/installing-rhel-on-adm64-intel-64-and-64-bit-arm
`x86_64` overall is supported on Fedora:
https://getfedora.org/en/coreos?stream=stable

max-cx:
I've opened a SSG issue: https://github.com/redhat-documentation/supplementary-style-guide/issues/171

rohennes:
@themr0c - Hi Fabrice, were you looking for somebody to work on removing this from our list of terms (as a good newbie task!). I can open a PR next week for that if that works?

themr0c:
That would be awesome!

rohennes:
[...]

PR #466, opened 2023-03-01
Move x86_64 to suggestion

aireilly:
x86_64 should not be flagged as an error. Demoting to suggestion.


aireilly:
Fixes https://github.com/redhat-documentation/vale-at-red-hat/issues/462

rohennes:
There was also #199 that referenced this question. I was also following that SSG conversation. It seems it is sometimes acceptable.

aireilly:
Ah didn't even know that one was there. Cool, let's get this merged then.

rohennes (approved):
Can possibly configure a bit more when we see the SSG resolution

PR #468, opened 2023-03-02
Updating workflows - save-state and set-output commands are deprecated

aireilly:
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


aireilly:
Just going to merge this, I will revert if there is an issue.

PR #477, opened 2023-03-09
"Amazon Web Services" should not flag as an error term

aireilly:
Remove the `webservices|web services|Web Services: Web services` rule altogether. It is misleading outside of JBoss EAP docs.

Context: 
* https://redhat-internal.slack.com/archives/C04HKR61MN1/p1678368726101069
* https://redhat-internal.slack.com/archives/C0218RXJK5E/p1678354141706229


rolfedh (approved):
LGTM

PR #479, opened 2023-03-10
Do not report errors for DM Multipath

jhradilek:
At the moment, Vale incorrectly reports the following error for any occurrence of "DM Multipath":

```
 61:1  error  Use 'Directory Manager' rather  RedHat.CaseSensitiveTerms 
              than 'DM'.
```
This is obviously problematic for Red Hat Enterprise Linux documentation that uses this term excessively. It is also potentially misleading because in this situation DM stands for "Device Mapper" so anyone following the advice would introduce a factual error in the documentation.

### DIFF
--- a/.vale/fixtures/RedHat/CaseSensitiveTerms/testvalid.adoc
+++ b/.vale/fixtures/RedHat/CaseSensitiveTerms/testvalid.adoc
@@ -58,6 +58,7 @@ DevOps
 Directory Manager
 Directory Server
 Disk Druid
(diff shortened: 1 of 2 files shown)


aireilly:
LGTM

aireilly:
Happy to merge?

jhradilek:
Thank you, @aireilly, for a lightning-fast review! I do not intend to make any more changes so unless I can explain or improve something, please merge when you are ready.

PR #480, opened 2023-03-13
removed webview/Webview from spelling errors

deerskindoll:
Since both "webview" and "Webview" are valid spellings, I removed the entry from SpellingErrors


aireilly (approved):
LGTM

PR #482, opened 2023-03-17
Recommend lower case for "Image Builder"

jhradilek:
The official position of Red Hat is to spell "image builder" with lower case letters to distinguish it from official products that use capitalization. This applies to formal documentation as well as blog posts and other documents. As this presents a change to how we used to refer to it, having Vale report it would be a great help.


aireilly (approved):
LGTM

PR #489, opened 2023-03-23
feat: don't report an error for 'APIs' (#489

### LINKED ISSUE #488: don't report an alert for 'APIs'
API is a common acronym that do'nt require a definition at firts use. It should be the same for it's plural form, APIs.

themr0c:
* fixes #488
* fixes also other rules that were raising false positive alerts on words in the `Definitions/testvalid.adoc` file

### DIFF
--- a/.vale/fixtures/RedHat/Definitions/testvalid.adoc
+++ b/.vale/fixtures/RedHat/Definitions/testvalid.adoc
@@ -1,6 +1,8 @@
+// suppress inspection "IncorrectFormatting" for whole file
 ACPI
 AIX
 API
+APIs
 ARM
 ASCII
 ASP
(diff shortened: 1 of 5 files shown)


.vale/fixtures/RedHat/Definitions/testvalid.adoc
```diff
@@ -1,6 +1,8 @@
+// suppress inspection "IncorrectFormatting" for whole file
```
aireilly, line 1: What does this mean?

themr0c: That's a comment that IntelliJ added to suppress one inspection and disable autoformatting.

Issue #517, opened 2023-04-28
Add template to manage the "flood of suggestions" issue and document it


rolfedh:
**Is your feature request related to a problem? Describe.**

Vale at Red Hat users express frustration with wading through the flood of repetitive suggestions. In some cases, they state that this experience discourages them from adopting the tool.

Although using the `--minAlertLevel=warnings` option mitigates this issue, it also hides potentially useful feedback. If we enable the `--minAlertLevel=warnings` option by default, what value are the suggestions? 

As a person developing a new style rule, the "flood of suggestions" issue and corresponding `--minAlertLevel=warnings` option complicate the decision of whether to make a rule a suggestion or a warning.

**Describe the solution you'd propose**

In https://github.com/errata-ai/vale/issues/614, JD Kato proposes using a template and provides an example template. I propose implementing and documenting use of this template. Can we distribute it along with the RedHat style?  

**Describe alternatives you've considered**

TBD

**Additional context**

TBD

PR #518, opened 2023-04-28
Add collate template

rolfedh:
Fixes https://github.com/redhat-documentation/vale-at-red-hat/issues/517

### DIFF
--- a/modules/user-guide/pages/collating-vale-output.adoc
+++ b/modules/user-guide/pages/collating-vale-output.adoc
@@ -1,0 +1,59 @@
+// Metadata for Antora
+:navtitle: Collating Vale output
+:keywords: collating, output
+:description: Collating the Vale CLI report results
+:page-aliases: end-user-guide:collating-vale-output.adoc
+// End of metadata for Antor
+
+:context: collating-vale-output
+:_module-type: CONCEPT
+[id="collating-vale-output_{context}"]
+= Collating Vale output
+
+You can use a template to collate the results from running `vale` on the command line. This collated output is non-repetitive and might be easier to use.
+
+The `RedHat` style includes a `collate-output.tmpl` template you can specify when you run the `vale` command. For example:
+
+----
+$ cd openshift-docs/
+$ vale --output='.vale/styles/RedHat/collate-output.tmpl' modules/about-crio.adoc
+----
+
+The resulting command output collates any repetitive the suggestions, warnings, or errors.
+
+.Example `vale` output from the `collate-output.tmpl` template
+----
+ modules/about-crio.adoc
+
+ 7:9,11:107  suggestion  Define acronyms and             RedHat.Definitions
+                         abbreviations (such as 'CRI')
+                         on first occurrence if they're
+                         likely to be unfamiliar.
+
+0 errors, 0 warnings and 2 suggestions in 1 file.
+----
+
+[NOTE]
+====
+When you use the `--output` option, other options such as `--minAlertLevel` might not work.
+====
+
+[id="understanding-vale-alert-levels_{context}"]
+== Understanding the `vale` output
+
+First column::
+A comma-separated list of each instance where the issue appears: `7:9,11:107`.
+
+Second column::
+The alert level, such as `error`, `suggestion`, or `warning`. See: xref:#understanding-vale-alert-levels_{context}[Understanding Vale alert levels]
+
+Third column::
+The rule message, such as: `Define acronyms and abbreviations (such as 'CRI') on first occurrence if they're likely to be unfamiliar.`
+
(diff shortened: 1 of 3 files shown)


rolfedh:
@aireilly  What would be the best location for this file to get distributed when users run `vale sync`?

modules/user-guide/pages/collating-vale-output.adoc
```diff
@@ -0,0 +1,70 @@
...
+:description: Collating the Vale CLI report results
+:page-aliases: end-user-guide:collating-vale-output.adoc
+// End of metadata for Antor
+
+:context: collating-vale-output
+:_module-type: CONCEPT
+[id="collating-vale-output_{context}"]
+= Collating Vale output
+
+As a technical writer, you can use a template to collate the results from running `vale` on the command line. This collated output is non-repetitive and might be easier to use.
```
aireilly, line 13: Is this "As a technical writer..." construction useful? Anyone can use the Vale at Red Hat style. 

```suggestion
You can use a template to collate the results from running `vale` on the command line. This collated output is non-repetitive and might be easier to use.
```

aireilly (commented):
Looks good and useful!

aireilly:
> @aireilly What would be the best location for this file to get distributed when users run `vale sync`?

So, to include this in the vale sync package, you would need to zip this alongside the yml in `.vale/styles/RedHat`. I'm not sure this is the right way to handle this though. Do we really want to include an optional output template alongside rules? We could provide a script to pull down the template maybe.

aireilly:
> > @aireilly What would be the best location for this file to get distributed when users run `vale sync`?
> 
> So, to include this in the vale sync package, you would need to zip this alongside the yml in `.vale/styles/RedHat`. I'm not sure this is the right way to handle this though. Do we really want to include an optional output template alongside rules? We could provide a script to pull down the template maybe.

On second thoughts, we already include an adoc file in the RH rules. I am OK to include this template there :)

modules/user-guide/pages/collating-vale-output.adoc
```diff
@@ -0,0 +1,70 @@
...
+
+The `RedHat` style includes a `collate-output.tmpl` template you can specify when you run the `vale` command:
+
+----
+$ cd openshift-docs/logging/troubleshooting/
+$ vale --output='~/styles/collate-output.tmpl' cluster-logging-troubleshooting-for-critical-alerts.adoc
```
aireilly, line 19:
```suggestion
$ vale --output='./.vale/styles/RedHat/collate-output.tmpl' logging/troubleshooting/cluster-logging-troubleshooting-for-critical-alerts.adoc
```

modules/user-guide/pages/collating-vale-output.adoc
```diff
@@ -0,0 +1,70 @@
...
+:_module-type: CONCEPT
+[id="collating-vale-output_{context}"]
+= Collating Vale output
+
+As a technical writer, you can use a template to collate the results from running `vale` on the command line. This collated output is non-repetitive and might be easier to use.
+
+The `RedHat` style includes a `collate-output.tmpl` template you can specify when you run the `vale` command:
+
+----
+$ cd openshift-docs/logging/troubleshooting/
```
aireilly, line 18:
```suggestion
$ cd openshift-docs/
```

modules/user-guide/pages/collating-vale-output.adoc
aireilly: The idea here is that we `vale sync` in `~/openshift-docs`, and then run Vale from the root project dir, this allows us to use the template using a relative path.

aireilly: depending on how you config Vale I guess

modules/user-guide/pages/collating-vale-output.adoc
rolfedh: Updated. Thanks.

modules/user-guide/pages/collating-vale-output.adoc
rolfedh: Updated. Thanks.

modules/user-guide/pages/collating-vale-output.adoc
rolfedh: Updated. Thanks.

modules/user-guide/pages/collating-vale-output.adoc
rolfedh: I'm not sure this works. Let's talk about it next week.

rolfedh:
@aireilly I've addressed the issues we discussed and verified the content. I think this is ready to merge.

PR #521, opened 2023-05-05
Fix callout rules

aireilly:
* Fixes implementation of MatchingNumberedCallouts. Previously, the rule simply counted the number of numbered callouts inside and outside the codeblock. This means that the rule would see no error in the following: 

```asciidoc
//vale-fixture
[source,ruby]
----
require 'frank' <1>
get '/hihi'<2><3>
----
<5> text
<6> More text
<7> More text again
``` 

The rule is now more robust - it reads every callout, and if a named callout is not found outside, it throws an error.


aireilly:
There is a problem with hard wrapped line: 
```asciidoc
<1> This provider name is prefixed to the returned user ID to form an identity
name.
<2> Controls how mappings are established between this provider's identities and `User` objects.
<3> List of attributes to use as the identity. First non-empty attribute is used. At least one attribute is required. If none of the listed attribute
have a value, authentication fails. Defined attributes are retrieved as raw,
allowing for binary values to be used.
<4> List of attributes to use as the email address. First non-empty
attribute is used.
<57> List of attributes to use as the display name. First non-empty
attribute is used.
<6> List of attributes to use as the preferred user name when provisioning a user for this identity. First non-empty attribute is used.
<7> Optional DN to use to bind during the search phase. Must be set if `bindPassword` is defined.
<8> Optional reference to an {product-title} `Secret` object containing the bind password. Must be set if `bindDN` is defined.
<9> Optional: Reference to an {product-title} `ConfigMap` object containing the PEM-encoded certificate authority bundle to use in validating server
certificates for the configured URL. Only used when `insecure` is `false`.
<10> When `true`, no TLS connection is made to the server. When `false`,
`ldaps://` URLs connect using TLS, and `ldap://` URLs are upgraded to TLS.
This must be set to `false` when `ldaps://` URLs are in use, as these
URLs always attempt to connect using TLS.
<11> An RFC 2255 URL which specifies the LDAP host and search parameters to use.
```

aireilly:
image

aireilly:
This should be ready for merge now. Remaining errors are ifdef codeblocks and legit errors.
[vale_errors_matching_callout.txt](https://github.com/redhat-documentation/vale-at-red-hat/files/11414792/vale_errors_matching_callout.txt)

PR #523, opened 2023-05-08
Adding OpenShiftAsciiDoc rule for hard wrapped text

aireilly:
Rule that flags when a hard-wrapped line is detected in an AsciiDoc file
image


rohennes:
This implementation looks good. Although I'm a bit confused about the reason behind it... no hard-wrapping is allowed between characters 65 and 100 on a line? But it's allowed outside of those characters?

Only found one false positive so far (I think), the ifdef line triggers is here:

````
ifdef::openshift-origin,openshift-enterprise,openshift-webscale[]
* Hybrid cloud deployments. You can deploy {product-title} clusters to a variety of public cloud platforms or in your data center.
endif::[]
````

aireilly:
> This implementation looks good. Although I'm a bit confused about the reason behind it... no hard-wrapping is allowed between characters 65 and 100 on a line? But it's allowed outside of those characters?
> 
> Only found one false positive so far (I think), the ifdef line triggers is here:
> 
> ```
> ifdef::openshift-origin,openshift-enterprise,openshift-webscale[]
> * Hybrid cloud deployments. You can deploy {product-title} clusters to a variety of public cloud platforms or in your data center.
> endif::[]
> ```



> ifdef::openshift-origin,openshift-enterprise,openshift-webscale[]
> * Hybrid cloud deployments. You can deploy {product-title} clusters to a variety of public cloud platforms or in your data center.
> endif::[]

Updated to account for if/endif lines

rohennes:
Or images, this triggers the rule:
`image::product-workflow-overview.png[High-level {product-title} flow]`

rohennes (approved):

PR #535, opened 2023-05-23
Remove Cloud > cloud term suggestion - it's not useful

aireilly:
Remove Cloud > cloud term suggestion - it's not useful


rohennes (approved):
Agreed, too many variations of Cloud to make it a solid rule.

PR #543, opened 2023-05-24
Adding Operator

aireilly:
Adding Operator to case sensitive words list. 

https://redhat-documentation.github.io/supplementary-style-guide/#operator


rohennes (approved):
We have a lot of hits for "operator" in the API specs. "Valid operators are In, NotIn, Exists and DoesNotExist." However, this is auto-generated content so writers won't be editing these modules. Therefore I think it's good as it will catch lowercase operator in the general docs
image

aireilly:
Removed in https://github.com/redhat-documentation/vale-at-red-hat/pull/544

PR #547, opened 2023-05-26
Adds VMware as pascal exception and allows "machine"

jeana-redhat:
Per discussion in https://github.com/openshift/openshift-docs/pull/60604#issue-1727661831

- VMware is an allowed word, but was tripping the pascal case rule. Added an exception.
- ISG says not to use "machine" but we use it (validly) all the time, and I don't think that's specifically just OCP docs. Removed it from the usage list.


aireilly:
Lgtm

PR #554, opened 2023-06-08
add rule for descriptive link text

neal-timpe:
Based on https://redhat-documentation.github.io/supplementary-style-guide/?topic=grammar-capitalization#cloud-services-links-hypertext

I added a rule that I hope flags empty link text and link text that only contains "click here" in a case insensitive way.

I also added pages for link text and images.

### DIFF
--- a/.vale/fixtures/AsciiDoc/LinkContainsLinkText/.vale.ini
+++ b/.vale/fixtures/AsciiDoc/LinkContainsLinkText/.vale.ini
@@ -1,0 +1,5 @@
+; Vale configuration file to test the `LinkContainsLinkText` rule
+StylesPath = ../../../styles
+MinAlertLevel = suggestion
+[*.adoc]
+AsciiDoc.LinkContainsLinkText = YES
--- a/.vale/fixtures/AsciiDoc/LinkContainsLinkText/testinvalid.adoc
+++ b/.vale/fixtures/AsciiDoc/LinkContainsLinkText/testinvalid.adoc
@@ -1,0 +1,14 @@
+//vale-fixture
+link:https://docs.aws.amazon.com/[] 
+
+link:https://docs.aws.amazon.com/[Click here]
+
+link:https://docs.aws.amazon.com/[click here]
+
+link:https://docs.aws.amazon.com/[here]
+
+link:https://docs.aws.amazon.com/[Here]
+
+link:https://docs.aws.amazon.com/[this]
+
+link:https://docs.aws.amazon.com/[This]
--- a/.vale/styles/AsciiDoc/LinkContainsLinkText.yml
+++ b/.vale/styles/AsciiDoc/LinkContainsLinkText.yml
@@ -1,0 +1,10 @@
+---
+extends: existence
+scope: raw
+level: warning
+ignorecase: true
+nonword: true
+link: https://redhat-documentation.github.io/supplementary-style-guide/#cloud-services-links-hypertext
+message: "Link is missing descriptive link text for accessibility."
+raw:
+  - 'link:.*(\[\]|\[click here\]|\[here\]|\[this\])'
--- a/modules/reference-guide/pages/links.adoc
+++ b/modules/reference-guide/pages/links.adoc
@@ -1,0 +1,10 @@
+:navtitle: Links
+:keywords: reference, rule, Links
+
+= Links
+
+Always include link text for URLs. The link text must be descriptive.
+
+.Additional resources
+
+* link:{ssg-url}?topic=grammar-capitalization#cloud-services-links-hypertext[{ssg} - Links and hypertext]
--- a/.vale/fixtures/AsciiDoc/ImageContainsAltText/.vale.ini
+++ b/.vale/fixtures/AsciiDoc/ImageContainsAltText/.vale.ini
@@ -1,4 +1,4 @@
-; Vale configuration file to test the `XrefContainsAnchorID` rule
+; Vale configuration file to test the `ImageContainsAltText` rule
 StylesPath = ../../../styles
 MinAlertLevel = suggestion
 [*.adoc]
--- a/.vale/fixtures/AsciiDoc/LinkContainsLinkText/testvalid.adoc
+++ b/.vale/fixtures/AsciiDoc/LinkContainsLinkText/testvalid.adoc
@@ -1,0 +1,2 @@
+//Correct usage of link text in links
+link:https://docs.aws.amazon.com/[AWS documentation] 
--- a/modules/reference-guide/nav.adoc
+++ b/modules/reference-guide/nav.adoc
@@ -9,6 +9,8 @@
 * xref:ellipses.adoc[]
 * xref:headingpunctuation.adoc[]
 * xref:headings.adoc[]
+//* xref:images.adoc[]
+//* xref:links.adoc[]
 * xref:oxfordcomma.adoc[]
 * xref:pascalcamelcase.adoc[]
 * xref:passivevoice.adoc[]
--- a/modules/reference-guide/pages/images.adoc
+++ b/modules/reference-guide/pages/images.adoc
@@ -1,0 +1,10 @@
+:navtitle: Images
+:keywords: reference, rule, Images
+
+= Images
+
+Use text equivalents for every diagram, image, or other non-text element.
+
+.Additional resources
+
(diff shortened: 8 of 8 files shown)


.vale/fixtures/AsciiDoc/LinkContainsLinkText/.vale.ini
```diff
@@ -0,0 +1,5 @@
+; Vale configuration file to test the `XrefContainsAnchorID` rule
```
aireilly, line 1:
```suggestion
; Vale configuration file to test the `LinkContainsLinkText` rule
```

.vale/fixtures/AsciiDoc/LinkContainsLinkText/testinvalid.adoc
```diff
@@ -0,0 +1,2 @@
+//vale-fixture
```
aireilly, line 1: Create a //vale-fixture for every entry in your rule please. That way we know the rule works for everything you want to catch

.vale/styles/AsciiDoc/LinkContainsLinkText.yml
```diff
@@ -0,0 +1,10 @@
+---
+extends: existence
+scope: raw
+level: warning
+link: https://redhat-documentation.github.io/supplementary-style-guide/#cloud-services-links-hypertext
+message: "Link is missing accessibility descriptive link text."
+raw:
+  - 'link:.*\[\]'
+  - 'link:.*\[[cC]lick here\]'
+
```
aireilly, line 10:
```suggestion

```

aireilly (changes requested):
Looks good! This will be a nice addition to the rule set :)

I pulled the PR locally and couldn't get the rule to work though. Something is not quite right.

.vale/styles/AsciiDoc/LinkContainsLinkText.yml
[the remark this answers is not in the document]
aireilly: Also, for some reason, the rule doesn't work when the two raw entries are present. when I update it like so, it does work: 
```yml
---
extends: existence
scope: raw
level: warning
link: https://redhat-documentation.github.io/supplementary-style-guide/#cloud-services-links-hypertext
message: "Link is missing accessibility descriptive link text."
raw:
  - 'link:.*\[\]'
```

rohennes:
Great work @neal-timpe ! This is a very useful. 

I wonder should could we add this to the [existing rule](https://github.com/redhat-documentation/vale-at-red-hat/blob/main/.vale/styles/AsciiDoc/ImageContainsAltText.yml) about alt text in images and create a more generic accessibility rule? 
So the regex would look like this: https://studio.vale.sh/s/e3d3e8259c271514938272107e51d1f4
You could rename the rule something more generic and update the rule and the test fixtures. WDYT?
cc @aireilly

aireilly:
> Great work @neal-timpe ! This is a very useful.
> 
> I wonder should could we add this to the [existing rule](https://github.com/redhat-documentation/vale-at-red-hat/blob/main/.vale/styles/AsciiDoc/ImageContainsAltText.yml) about alt text in images and create a more generic accessibility rule? So the regex would look like this: https://studio.vale.sh/s/e3d3e8259c271514938272107e51d1f4 You could rename the rule something more generic and update the rule and the test fixtures. WDYT? cc @aireilly
> 
> FYI - there seems to be something weird with using tokens for this rule. I've opened an [issue](https://github.com/errata-ai/vale/issues/643) on the Vale repo.

Sounds good to me! A generic `AltText` rule would be good.

rohennes:
https://github.com/errata-ai/vale/issues/643#issuecomment-1585131513 - Just needed to add `nonword: true` for the rule  to work with tokens. So part an AltText rule might look like this: https://studio.vale.sh/s/bfecfefeab7169a4cd15aa9f999fec5c

neal-timpe:
@rohennes @aireilly Thanks for the feedback! I hear what you're saying about combining them. However, I have a plan for a couple more accessibility rules. I figured this was containable and I wanted to make sure I could write one coherent one before I tried to write more than one. :) 

What do you think? Should we still combine them if there are going to be a couple others?

I'll address these other comments and squash while we discuss.

.vale/styles/AsciiDoc/LinkContainsLinkText.yml
neal-timpe: Is this suggested to remove a trailing newline?

.vale/styles/AsciiDoc/LinkContainsLinkText.yml
```diff
@@ -0,0 +1,10 @@
+---
+extends: existence
+scope: raw
+level: warning
+link: https://redhat-documentation.github.io/supplementary-style-guide/#cloud-services-links-hypertext
+message: "Link is missing accessibility descriptive link text."
+raw:
```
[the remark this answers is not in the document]
neal-timpe: Instead of the tokens, I took Ronan's example. https://studio.vale.sh/s/e3d3e8259c271514938272107e51d1f4

.vale/styles/AsciiDoc/LinkContainsLinkText.yml
aireilly: You had 2 spaces and a new line :sweat_smile:

A nit - my yaml linter reported it. Usually not a problem.

aireilly:
> @rohennes @aireilly Thanks for the feedback! I hear what you're saying about combining them. However, I have a plan for a couple more accessibility rules. I figured this was containable and I wanted to make sure I could write one coherent rule before I tried to write more than one. :)
> 
> What do you think? Should we still combine them if there are going to be a couple others?
> 
> I'll address these other comments while we discuss.

Are the planned rules for alt text for various elements? I think that a single alt text rule makes sense tbh. But I can be convinced otherwise :)

neal-timpe:
@aireilly No, the rules I have in mind are non-alt text. For example, one idea is a warning for direction words and colors. 

From my perspective, I think that the link text and the image alt text are different enough to have their own rules. Link text is there to provide context. Image alt text must be equivalent to the image's purpose. I think in my next PR, I might make a stab at updating the image alt text message to make that more specific guidance in a short a space as possible.

modules/reference-guide/pages/links.adoc
```diff
@@ -0,0 +1,10 @@
+:navtitle: Links
+:keywords: reference, rule, Links
+
+= Links
```
aireilly, line 4: Side-note - none of the AsciiDoc rules are documented as yet. The reference guide has Red Hat rules only. Might be better to remove these ref guide entries for now until such time as we have an proper AsciiDoc rules section.

.vale/styles/AsciiDoc/LinkContainsLinkText.yml
```diff
@@ -0,0 +1,9 @@
+---
+extends: existence
+scope: raw
+level: warning
+ignorecase: true
```
aireilly, line 5:
```suggestion
ignorecase: true
nonword: true
```

Think we should also set `nonword: true`. hmm.

aireilly: @rohennes ?

rohennes: I think so. Otherwise the angle brackets can be excluded from regex

https://github.com/errata-ai/vale/issues/643#issuecomment-1585131513

rohennes:
> @rohennes @aireilly Thanks for the feedback! I hear what you're saying about combining them. However, I have a plan for a couple more accessibility rules. I figured this was containable and I wanted to make sure I could write one coherent rule before I tried to write more than one. :)
> 
> What do you think? Should we still combine them if there are going to be a couple others?
> 
> I'll address these other comments while we discuss.

Understood and I'm good with your suggestion for multiple rules. These will be great contributions!!

modules/reference-guide/pages/links.adoc
neal-timpe: Ok, I commented links.adoc and images.adoc out in the nav.adoc file so we don't have to do it again later.

aireilly:
Looking pretty good @neal-timpe Can we squash the commits before merge?

neal-timpe:
@aireilly Squashed. :)

aireilly (approved):
/lgtm

PR #572, opened 2023-06-20
Accomodating `==` format headings for additional resources rule

rohennes:
See here: https://github.com/openshift/openshift-docs/blob/main/contributing_to_docs/doc_guidelines.adoc#additional-resources-sections

This updates the Additional resources rule for assemblies to ensure it catches `== Additional resources` and `.Additional resources`. Dot formatting can be used in modules or assemblies. == format is used in assemblies only.


aireilly (approved):
/lgtm

Issue #530, opened 2023-05-11
Create a Vale rule that checks for official and valid word usage


abhatt-rh:
Create a Vale rule that checks for official and valid word usage across docs. The rule (or rules) could check for the following:

- Red Hat official Product names
- IBM Style Word Usage
- Red Hat supplementary style guide for product documentation

PR #578, opened 2023-07-10
Add new Vale rule to catch instances of "Red Hat Java"

michelle-purcell:
Enhances vale-at-red-hat rules as per recent CCS communication requesting that all instances of `Red Hat Java` are replaced with `Red Hat OpenJDK` or the full product name `Red Hat build of OpenJDK`.

Extract: 
"...always mention Red distribution of JDK as Red Hat OpenJDK.
Never use Red Hat Java in any content"

_**Note:** Not sure if the Red Hat `CaseSensitiveTerms.yml` rule is the right place for this but as per other similar Red Hat product name rules live here... I think these fall under a rule for product names as per @abhatt-rh's enhancement GH issue #530._


rohennes (approved):
LGTM

PR #579, opened 2023-07-13
accessible visual language

neal-timpe:
Create a warning for only using visual words.
https://issues.redhat.com/browse/CCSENABLE-501


.vale/styles/RedHat/VisualLanguage.yml
```diff
@@ -0,0 +1,22 @@
...
+extends: existence
+ignorecase: true
+level: suggestion
+link: https://redhat-documentation.github.io/vale-at-red-hat/docs/main/reference-guide/visuallanguage/
+message: "It is acceptable to use visual styling in documentation, but do not rely solely on visual aspects to convey meaning."
+tokens:
+  - above
+  - below
+  - right
+  - left
```
aireilly, line 11: Can we rework this list to include appropriate quickfix suggestions as outlined in the SSG for directional terms?

e.g., 
```
above: previous|preceding
below: following
```

The rule would then need to be a https://vale.sh/docs/topics/styles/#substitution rule. This may not be suitable for all these terms though.

neal-timpe: I thought about that, but I think that in a lot of cases the rewriter will be more complicated than switching the word out. For example, if we're talking about `the menu on the left side of the page`, which I see quite often, substituting left for something won't work. The best practice is to name the menu or ui control. And getting rid of left likely means rewriting the sentence.

It's also allowable to use visual language if you also provide a non visual way to find it.

What do you think?

aireilly: Looking at https://www.ibm.com/docs/en/ibm-style?topic=word-usage#section-use-with-caution-left-adjective, the rule as you have it here would flag left even when it was used correctly, "The list of tables is displayed in the Table List pane on the left side of the window".

I'm not sure how we could reduce false positives in this case with simple regex. I wonder could a sequence rule be useful? https://vale.sh/docs/topics/styles/#sequence

neal-timpe: I think you're right. I'm struggling with the sequence rule because the pattern that comes before or after could be anything. 

I'm not sure there's a way to reduce false positives in this case. And I was aware of that when I started this rule. It is important. What if we made it a warning, for example, if you're going to use these visual words, make sure you have it another way too?

aireilly: Not quite what you are looking for, but here are swap terms (bad:good) lifted from the ISG: 

```
bottom left: lower left
bottom right: lower right
left most: leftmost
left-hand: left
left-justified: left-aligned
left-justify: left-align
left-most: leftmost
right double-click: double right-click
right-hand: right
right-justified: right-aligned
right-justify: right-align
right-most: rightmost
```

aireilly: Another thought - grep search across a few repos, see if there are common misused terms we can flag.

aireilly: more suggestions:

below: earlier|later|following
above: later|previous|preceding

neal-timpe:
These are good suggestions. Sorry I was on pto for a couple of weeks. I didn't meant to ignore this. :)

aireilly:
Note this new rule could also serve as an example for this PR: 
https://github.com/redhat-documentation/vale-at-red-hat/blob/main/.vale/styles/RedHat/DoNotUseTerms.yml

aireilly:
Here is another approach. We have a new Error rule that provides context to the user about error terms. See https://github.com/redhat-documentation/vale-at-red-hat/blob/main/.vale/styles/RedHat/DoNotUseTerms.yml

Could work here too. For example,

```yaml
---
extends: substitution
ignorecase: true
level: suggestion
link: https://redhat-documentation.github.io/vale-at-red-hat/docs/main/reference-guide/visuallanguage/
message: "%s"
swap:
  # Error messages must be single quoted.
  above: 'Try not to use "above". Visual language may be inaccessible to some readers. This can hinder comprehension and inclusivity in your content.'
#etc
```

neal-timpe:
That actually sounds like what I want to do. There are some items for which I have a swap. But there are others, like colors, where writers really need to be warned not to use them in certain cases.

.vale/styles/RedHat/DoNotUseTerms.yml
```diff
@@ -17,6 +17,7 @@ swap:
   devs?: 'Do not use "dev" as a synonym for development or developer.'
   foo: 'Do not use "foo". This term is technical jargon in code and as shorthand for fubar, an acronym of profanity in code.'
   fubar: 'Do not use "fubar". This term is an acronym of a profanity that is sometimes used by developers in code.'
+  green|yellow|red|blue: 'Do not use visual language, like color, to describe items in the us unless it is also described non-visually, like with the name of the item.'
```
aireilly, line 20: Not sure about this as a "Do Not Use" term. For one thing, this will throw an error at every instance of "Red Hat" in source. We've also had strong feedback about noise in the Vale rules and we've had some teams actively turn away from our Vale rules because of this. My inclination is to leave out an error term if there is ambiguity as to its correctness in a given context.  

ISG says: "[Don't] Use color as the only way to convey meaning; information that is conveyed by color should also be evident when it's rendered without color."

This does not say that we shouldn't use colors in our written content, just that we shouldn't use it as the _only_ way to convey meaning.

https://www.ibm.com/docs/en/ibm-style?topic=medium-accessibility

neal-timpe: Ok, finally got around to this. I removed the new rule that I was trying to make called visual language, and put all the  items into already existing rules. I also got rid of the admonition against Red.

neal-timpe:
@aireilly I think this one i ready for another look. If you're okay with it, I'm happy to squash.

.vale/styles/RedHat/DoNotUseTerms.yml
```diff
@@ -16,6 +16,7 @@ swap:
   debuggable: 'Do not use "debuggable". Rephrase the sentence to use the verb or noun debug. For example, change rebuild the debuggable version to rebuild the version that can be debugged.'
   foo: 'Do not use "foo". This term is technical jargon in code and as shorthand for fubar, an acronym of profanity in code.'
   fubar: 'Do not use "fubar". This term is an acronym of a profanity that is sometimes used by developers in code.'
+  green|yellow|blue: 'Do not use visual language, like color, to describe items in the us unless it is also described non-visually, like with the name of the item.'
```
aireilly, line 19:
```suggestion
  \bred\b(?!-? *hat)|orange|yellow|green|blue|indigo|violet: 'Do not use colors to describe something unless it is also described non-visually, for example, with the name of the item.'
```

I tested this regex and it won't catch "redhat", "red hat" or "red-hat". 

Also added ROYGBIV colors because why not - unless there is a reason to only flag green/yellow/blue?

aireilly: Looks good otherwise! Will be good to get this in

neal-timpe:
@aireilly Added your last update and squashed.

aireilly (approved):
/lgtm 

:1st_place_medal: thank you for hanging in there Neal! The effort is much appreciated.

PR #580, opened 2023-07-14
Follow on rule enhancement to #578

michelle-purcell:
Further enhances the previously merged rule https://github.com/redhat-documentation/vale-at-red-hat/pull/578 following further Red Hat SME discussion about the guideline.

See also related Red Hat style PR [362](https://github.com/redhat-documentation/supplementary-style-guide/pull/362).

### DIFF
--- a/.vale/styles/RedHat/TermsErrors.yml
+++ b/.vale/styles/RedHat/TermsErrors.yml
@@ -244,6 +244,8 @@ swap:
   pseudo ops|pseudoops: pseudo-ops
   pull-down: pulldown
   recognise: recognize
+  Red Hat OpenJDK|RHOJDK|Red Hat Open Java Development Kit: Red Hat build of OpenJDK
+  Red Hat Java: Red Hat build of OpenJDK
   regex: regular expression
   remote-access server: remote access server
   remote-access: remote access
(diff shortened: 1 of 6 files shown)


.vale/styles/RedHat/TermsErrors.yml
```diff
@@ -244,6 +244,8 @@ swap:
   pseudo ops|pseudoops: pseudo-ops
   pull-down: pulldown
   recognise: recognize
+  Red Hat OpenJDK: Red Hat build of OpenJDK
```
rohennes, line 247: I see RHOJDK is also disallowed in the SSG PR. Is it something to include as well here? Or no worries if not.

.vale/styles/RedHat/TermsErrors.yml
```diff
@@ -244,6 +244,8 @@ swap:
...
+  Red Hat Java|Red Hat java: Red Hat build of OpenJDK
```
rohennes, line 248: `ignorecase: true` - is set in this rule so I think Red Hat Java is all that's required. It will ignore case so upper and lower cases are found.

.vale/styles/RedHat/TermsErrors.yml
michelle-purcell: @rohennes - Good catch. Yes, that would make perfect sense  +1. Thank you :green_heart:

.vale/styles/RedHat/TermsErrors.yml
michelle-purcell: @rohennes - Good point. No need to call it out so I removed it. Sorry bit rusty updating rules! Thanks for flagging.

rohennes (approved):
lgtm - thanks Michelle!

PR #582, opened 2023-07-21
Add execptions to RedHat.Slash style rule.

emteelb:
- TCP/IP
- <1-4 digits>/<1-4 digits>, that likely represent a date or numerical expression, outside the scope of this rule.


aireilly (approved):
lgtm

PR #539, opened 2023-05-23
WIP - Opinionated clean up of Vale terms

aireilly:
1. Removes all terms that have ambiguous meaning depending on context. Vale can't know the context, so these rules are not helpful.
2. Removes all extraneous text from code completion replacement suggestions.
3. Retires the usage rules in favour of an ISG rule that suggests alternative terms
4. Fixes https://github.com/redhat-documentation/vale-at-red-hat/issues/548

PR #584, opened 2023-07-26
Clean up suggestion terms

aireilly:
This PR cleans up and consolidates Suggestion terms. Where there is an ambiguity for a term, I have removed the term. All terms have correct quick fix options with no extraneous text. 

This PR is a more manageable piece of this larger PR: https://github.com/redhat-documentation/vale-at-red-hat/pull/539


aireilly:
fixed in https://github.com/redhat-documentation/vale-at-red-hat/pull/585

PR #585, opened 2023-07-26
Retire Usage rule and update Error rule

aireilly:
The current usage rule is out of date, has ambiguous terms guidance and is generally not very helpful.

This update removes the Usage rule completely and instead adds any ISG terms that have clear "Do not use" guidance and definitive suggested alternatives to the Error rule. All ambiguous terms or terms without suggested alternatives are removed.

Terms are pulled directly from the ISG word usage list: https://www.ibm.com/docs/en/ibm-style?topic=word-usage.

This PR is a more manageable piece of this larger PR: https://github.com/redhat-documentation/vale-at-red-hat/pull/539

DONE: ~move do-not-use terms without guidance to https://github.com/redhat-documentation/vale-at-red-hat/pull/591~


aireilly:
These terms could be added to this rule: 

```
  - as well as: and
  - below: earlier|later|following
  - above: later|previous|preceding
  - carry out: test|run
  - choose|check: select
  - consumes?: use|uses
  - uncompress: decompress
  - either: each
  - execute: run|issue|start|enter
  - fatal: unrecoverable
  - higher: later
  - hit: press|type
  - illegal: invalid|not allowed|incorrect
  - ingests?: import|load|obtain|process
  - invoke: start|call
  - kill: end|stop
  - legacy: existing|traditional|established|classic|earlier|previous
  - like: such as
  - since: because
  - so: so that|therefore
  - so long as: if|when
  - translate: convert:transform
  - demount: unmount|dismount
  - updatable: can be updated|can be changed
```

.vale/styles/RedHat/Usage.yml
```diff
@@ -1,151 +1,357 @@
...
+  syncpoint: sync point
+  system engineer: systems engineer
+  tablespace: table space
+  tap on: tap
+  TCP/IP network: Internet Protocol network|IP network
+  teleprocessing line: telecommunication line|transmission line
+  testcase: test case
+  third-party software vendor: independent software vendor
+  thumb drive: USB flash drive
+  time box: timebox
```
abhatt-rh, line 278: timeframe: time frame

.vale/styles/RedHat/Usage.yml
```diff
@@ -1,151 +1,357 @@
...
+  time line: timeline
+  time stamp: timestamp
+  time-out: timeout|time out
+  timeslice: time slice
+  timezone: time zone
+  tool bar: toolbar
```
abhatt-rh, line 284: toggle(?!on| off): toggle

.vale/styles/RedHat/Usage.yml
```diff
@@ -1,151 +1,357 @@
...
+  tool box: toolbox
+  tool kit: toolkit
+  tool tip: tooltip
+  ToolTip: tooltip
+  totalled: totaled
+  touch base: contact|communicate
+  trade-off|trade off: tradeoff
+  trouble shoot: troubleshoot
```
abhatt-rh, line 292: trust store: truststore

.vale/styles/RedHat/Usage.yml
```diff
@@ -1,151 +1,357 @@
...
+  try and: try to
+  twisty: twistie
+  TY-RAP: cable tie
+  un-install: uninstall
+  unallocate: deallocate
+  uncompilation: decompilation
```
abhatt-rh, line 298: uncheck: clear|deselect

.vale/styles/RedHat/Usage.yml
```diff
@@ -1,151 +1,357 @@
...
+  uncompile: decompile
+  underbar: underscore
+  unencrypt: decrypt
```
abhatt-rh, line 301: Unix|unix: UNIX
unjar: extract|decompress
unzip: extract|decompress
unselect: clear|deselect
U.K.:UK
U.S.|US
utilize: use

.vale/styles/RedHat/Usage.yml
```diff
@@ -1,151 +1,357 @@
...
+  upgradable: upgradeable
+  USB thumb drive: USB flash drive
+  useable: usable
+  vanilla: un-customized
```
abhatt-rh, line 305: Not sure about _un-customized_  based on the description [here](https://www.ibm.com/docs/en/ibm-style?topic=word-usage#do-not-use-vanilla-adjective-noun)
```suggestion
  vanilla:  not customized
```

.vale/styles/RedHat/Usage.yml
```diff
@@ -1,151 +1,357 @@
...
+  virtual DOS machine: DOS session
+  virtual-equals-real: nonpageable
+  virtual-equals-virtual: pageable
```
abhatt-rh, line 308: vs.: versus|compared to

.vale/styles/RedHat/Usage.yml
```diff
@@ -1,151 +1,357 @@
...
+  wait condition: wait state
+  war room: center of operations|command center|control room
+  warmstart|warm boot: warm start
+  web cast: webcast
+  web site: website
+  webmistress|web master: webmaster
+  while: although|though
```
abhatt-rh, line 315: whilst: while

.vale/styles/RedHat/Usage.yml
```diff
@@ -1,151 +1,357 @@
...
+  whitepaper: white paper
+  WiFi|Wifi|wi-fi: Wi-Fi
+  wild card: wildcard
+  wiring closet: telecommunications closet
+  work flow: workflow
+  work group: workgroup
+  work load: workload
+  work space: workspace
+  work station: workstation
+  XSA: extended subarea addressing
```
abhatt-rh, line 327: world wide|world-wide: worldwide

abhatt-rh (commented):
Huge thank to you @aireilly for this one. It's going to be super helpful for all of us!! 

Well, I added a bunch of terms. I started with updating both the `testvalid` and `testinvalid` docs, but later I felt it might make sense for me to just add them after this PR is merged rather than adding them as comments.

Further, there will be a huge overlap between the
[TermsErrors.yml](https://github.com/redhat-documentation/vale-at-red-hat/blob/main/.vale/styles/RedHat/TermsErrors.yml) rules and this one (Usage.yml). Do we plan to consolidate or clarify what's the difference between the two? We also have the do-not-use guidance from our SSG (I understand that it's addressed in #591).  

Let me know if something needs clarification or if I can help out.

aireilly:
@jhradilek Would appreciate your eyes on this. have been working on an off for the last few months on this one. Think it's nearly ready. I want to throw out the Usage rule and consolidate the TermsErrors rule. 

Running this to test (Pasting PR version TermsErrors.yml into the local .vale folder):
```
┌───── ~/rhel-8-docs/rhel-9 (master)*
$ vale ./modules --ext='.adoc' --output=line  --filter='.Name == "RedHat.TermsErrors"' --minAlertLevel=error > /home/aireilly/vale_test_updated_errors.txt
```
I get the following errors: http://pastebin.test.redhat.com/1108678

I think these are legitimate errors. WDYT?

jhradilek:
Hi @aireilly, I looked over your list and most of these are legitimate, I however found a few irregularities.

In the Word Usage section of the online version of the IBM Style Guide, the following are listed as warnings and advised to be used with caution in the specified context. I would have expected these in `RedHat.TermsWarnings`, not `RedHat.TermsErrors`:

- downgrade
- execute
- launch
- terminate

There may be more but these four jumped at me in your output file.

There are at least two words that recommend a replacement in direct contradiction with the IBM Style Guide:

- `Use 'turn off' rather than 'Power off'`: The IBM Style Guide only mentions "power off" in the guidance for "power down" and says the following: "Do not use. Use power off or turn off." This makes "power off" perfectly valid. The Red Hat Supplementary Style Guide does not have an opinion but happily uses "power off" in examples.
- `Use 'adviser' rather than 'Advisor'`: The IBM Style Guide says the exact opposite: "advisor - Write as shown: not adviser." and "adviser - Do not use. Use advisor."

I would also recommend removing `Subversion` or at least making that rule case sensitive as Apache Subversion (SVN) is still a common occurrence in technical documentation and thus not an error.

aireilly:
Thanks @jhradilek Re: Advisor, I see this in https://www.ibm.com/docs/en/ibm-style?topic=word-usage:  
image

jhradilek:
Hi @aireilly, thank you, that is consistent with what I quoted and I see it is correct in your branch so please ignore that comment. I was reviewing both pastebins for comparison and this is visible in the second one that uses rules from `main`.

aireilly:
Actually a large number of these need to move to the Warning rule.

aireilly:
This is looking pretty good now IMO (RHEL 9 docs errors with latest ErrorTerms rule): http://pastebin.test.redhat.com/1108758

jhradilek:
Great job, @aireilly, the updated error list looks pretty good indeed. I have only a few minor comments:

- `Use 'later' rather than 'higher'.` This should be a warning as the IBM Style Guide only restricts the use of the word when describing software versions. Some of the reports in your output are of course correct, but others complain about sentences like "value higher than the total number" which I think are valid. The suggestion does not make sense in this context.
- `Use 'uses' rather than 'consume'` and `Use 'use' rather than 'consumes'.` I believe these are flipped but should also be warnings as the IBM Style Guide explicitly states: "Consume is acceptable when it is part of the existing product terminology." Some of the reports in your output are correct, but others are related to Subscription Manager which even has the `consume` subcommand.
- `Use 'select' rather than 'Choose'.` This should be a warning as the IBM Style Guide only restricts the use of the word when referring to actions performed on interface element. Choosing a username or an installation method are valid uses for that word.
- `Use 'convert:transform' rather than 'translate'.` The one report in your output is right and the word should be fixed, but I still believe this should be a or a warning as the IBM Style Guide only restricts the use of the word when referring to data. In other parts of our documentation, you would find phrases like "The text of an application that cannot be translated into the selected language remains in US English" that are correct. Additionally, I would recommend adding rules for other forms of the word like "translates" or "translated" which are also present but currently unchecked.
- `Use 'executable program' rather than 'executable'.` This can stay as an error but I think we need to expand the regular expression to account for other valid uses of the word like "executable file" or "executable files".

aireilly:
Getting pretty close... http://pastebin.test.redhat.com/1108879

aireilly:
Something is screwy in the Workflows.

aireilly:
> I can still see "choose" reported as an error but otherwise looks good to me!

Oops! removed.

aireilly:
Merging this. We can fix remaining errors as we find them.

PR #590, opened 2023-07-27
Exclude snippet ranges from includes in assemblies

### LINKED ISSUE #532: OpenShiftAsciiDoc should disallow snippets that fetch certain lines only
In OpenShift, they want to disallow snippets that grab a few lines only. It's valid asciidoc but breaks the build. Create OpenShiftAsciiDoc rule to catch it.

https://github.com/openshift/openshift-docs/pull/60315

https://docs.asciidoctor.org/asciidoc/latest/directives/include-lines/#specifying-line-ranges

rohennes:
Fixes #532 

In the OpenShift repo, we should disallow this rule in the modules config file? We have a rule that flags any include in modules already. Probably don't need this too?


aireilly:
lgtm generally. One comment though - this should also apply to modules I think. You could include a snippet and specify a line range for example.

.vale/styles/OpenShiftAsciiDoc/NoSnippetRangesInAssemblies.yml
```diff
@@ -0,0 +1,8 @@
+---
+extends: existence
+scope: raw
+level: error
+link: https://github.com/openshift/openshift-docs/blob/main/contributing_to_docs/doc_guidelines.adoc#links-hyperlinks-and-cross-references
+message: "You can only nest snippets or GitHub raw user content in modules."
```
aireilly, line 6:
```suggestion
message: "Do not use line ranges for included files."
```

.vale/styles/OpenShiftAsciiDoc/NoSnippetRangesInAssemblies.yml
```diff
@@ -0,0 +1,8 @@
+---
+extends: existence
+scope: raw
+level: error
+link: https://github.com/openshift/openshift-docs/blob/main/contributing_to_docs/doc_guidelines.adoc#links-hyperlinks-and-cross-references
```
aireilly, line 5:
```suggestion
link: https://github.com/openshift/openshift-docs/blob/main/contributing_to_docs/doc_guidelines.adoc#assembly-file-metadata
```

aireilly (approved):
lgtm

PR #591, opened 2023-07-27
Adding a "do not use" terms rule

aireilly:
Adding a "do not use" terms rule. Instead of suggesting text replacement, the swap field is used to provide inline context to the user regarding the specific term.

```terminal
$ vale .

 testinvalid.adoc
 1:1  suggestion  Do not use "please". Terms of   RedHat.DoNotUseTerms 
                  politeness are superfluous,                          
                  convey the wrong tone for                            
                  technical material, and are                          
                  not regarded the same way in                         
                  all cultures.                                        
```

image

### DIFF
--- a/.vale/fixtures/RedHat/DoNotUseTerms/testinvalid.adoc
+++ b/.vale/fixtures/RedHat/DoNotUseTerms/testinvalid.adoc
@@ -1,0 +1,28 @@
+ACK
+basically
+congratulations
+daughterboard
+destroy
+devs
+domestic
+foo
+foreign
+fubar
+future-proof
+geo
+geography
+kerberize
+kerberized
+native interface
+out of the box
+out-of-the-box
+overhead
+please
+quiescent
+resides
+respective
+respectively
+time-tested
+we
+debuggable
+and/or
--- a/.vale/fixtures/RedHat/DoNotUseTerms/.vale.ini
+++ b/.vale/fixtures/RedHat/DoNotUseTerms/.vale.ini
@@ -1,0 +1,5 @@
+; Vale configuration file to test the `DoNotUseTerms` rule
+StylesPath = ../../../styles
+MinAlertLevel = suggestion
+[*.adoc]
+RedHat.DoNotUseTerms = YES
--- a/.vale/fixtures/RedHat/DoNotUseTerms/testvalid.adoc
+++ b/.vale/fixtures/RedHat/DoNotUseTerms/testvalid.adoc
@@ -1,0 +1,2 @@
+hash sign
+ACK flag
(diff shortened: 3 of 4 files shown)


rolfedh (approved):
Excellent work, @aireilly. Well-conceived and implemented.

aireilly:
Thanks @rolfedh my only concern would be these suggestions showing for legit uses. I might trim down the list for any terms that require context.

.vale/fixtures/RedHat/DoNotUseTerms/testinvalid.adoc
```diff
@@ -0,0 +1,29 @@
...
+destroy
+dev
+domestic
+foo
+foreign
+fubar
+future-proof
+geo
+geography
+latest
```
kcarmichael08, line 12: Hi, I was wondering, does adding these terms mean that Vale will fail on them? This is the only one that I question, because I think "install the latest version of the software" might be a fairly common directive and one where context matters. I'm unclear if failing Vale would prevent the build from happening. (ACS has some exceptions to Vale so we have it set up to allow the build, but I wasn't sure about OCP.)

aireilly: Agree I'll remove this thanks

ndeevy:
ahh it's merged, happy days @aireilly  :-] and it looks like you made the list way better, huzzah!

I had started a review but I'll just throw the notes here in case you want to consider them for a future update...

**geo/geography**
Because this is one of the more 'ambiguously incorrect' ones in the list, I wonder if it needs inclusion at all. After a search on the portal I didn't see lots of instances of this word being used incorrectly in the docs. Do we have some strong evidence it needs flagging through Vale? If not then maybe it'd be safe enough to trim from the list. 


**foreign/domestic**
Another 'ambiguously incorrect' one and after a cursory portal search, the hits that come up in docs are more like `foreign key`, `foreign hypervisor` etc., which aren't the incorrect usages that this entry is targeting. That means it's likely to throw up errors for those things instead. Maybe this one could go too if you want more strictly incorrect words. 

**destroy, kerberize, daughterboard** just querying these because I'm wondering are they used incorrectly in the docs enough to warrant inclusion? I found kerberize in KCS but not docs really, daughterboard I didn't find and destroy can be kind of ambiguously used. If you're looking to trim the list maybe consider these three as options for the cutting room floor.

Then I thought of a couple of unambiguous options you could consider adding, if you haven't already:
- **prior to**
- **i.e.**
- **etc.** 
- **e.g.**
 
Thanks

aireilly:
@ndeevy thanks I made thesae changes in https://github.com/redhat-documentation/vale-at-red-hat/pull/620

aireilly:
We already catch prior to, "i.e.", "etc", and "e.g.". I'll "Prior to" thanks.

PR #600, opened 2023-08-01
Add fixes to reduce spelling rule false positives

michelle-purcell:
Fixes:

- False positive spelling warnings on Java-related software services/components/products/functions and some acceptable widely used programmatic terms encountered when running vale on Red Hat build of Quarkus docs.
- Removed some duplicate entries in spelling yml.

_Related heading rule enhancements to follow after this is merged._

### DIFF
--- a/.vale/styles/RedHat/Spelling.yml
+++ b/.vale/styles/RedHat/Spelling.yml
@@ -23,13 +23,18 @@ filters:
   - "[bB]reakpoints"
   - "[cC]he"
   - "[cC]lassloading"
+  - "[cC]lasspath"
   - "[cC]olocate"
   - "[cC]onfig"
+  - "[cC]ustomizer"
+  - "[dD]atasource"
   - "[dD]eclaratively"
   - "[dD]ecompiler"
   - "[dD]eserialization"
+  - "[dD]eserialized"
   - "[dD]esynchronize"
   - "[dD]esynchronized"
+  - "[dD]ev"
   - "[dD]ev[wW]orkspace"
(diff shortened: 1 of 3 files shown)


aireilly:
Some errors to fix, but this looks like a very useful update!

.vale/styles/RedHat/Spelling.yml
```diff
@@ -384,9 +407,8 @@ filters:
   - Traefik
   - Uber
   - UIDs
-  - URIs?
-  - url
-  - URLs?
+  - URI
+  - URL
```
aireilly, line 411: "URLs" and "URIs" acceptable usages aren't they?

michelle-purcell: @aireilly Just to clarify, I removed the plural instances as those are not needed. The expression caters for plural instances of the listed term.
In my tests against this PR, `URLs` and `URIs` are not flagged with a spelling warning.

Also, I did wonder why several items have a question mark against them. Is that a workaround for an issue or a tag to investigate those terms?

michelle-purcell: @aireilly  Fixed build errors!

aireilly: the question mark is regex - it means, "one or more of the preceding character".

michelle-purcell: @aireilly - Doh :face_in_clouds: Thanks for reminding me! I restored those items.

aireilly (approved):
lgtm

PR #601, opened 2023-08-01
adding hunspell dictionary

aireilly:
Need to figure out how this would be delivered. Can we put it inside the vale sync archive? Or else will need another dictionary install step.

Vale uses an en_US dic by default. Are there other dictionaries? 

```
.vale/styles/RedHat/Spelling.yml:1:E201:open /__w/vale-at-red-hat/dictionaries/en_US.dic: no such file or directory
.vale/styles/RedHat/Spelling.yml:1:E201:open /__w/vale-at-red-hat/dictionaries/en_US.dic: no such file or directory
```


aireilly:
closing - may investigate further at a later point.

PR #602, opened 2023-08-01
Fix false positive churn flagged by the Headings and CaseSenstive rules

michelle-purcell:
This PR reduces false positives that come up regularly in Quarkus doc content.


aireilly (approved):
lgtm

michelle-purcell:
Thanks @aireilly.

PR #605, opened 2023-08-11
Do not suggest to replace "laptop" with "notebook"

jhradilek:
The current version of the IBM Style Guide does not prohibit the use of the word "laptop" in any way and recommends it to refer to mobile computer that is not a handheld device. On the contrary, the IBM Style Guide recommends to use the word "notebook" to refer only to an interface element that resembles a physical notebook with pages and tabs.

I am therefore proposing to remove this rule as it goes against what the style guide recommends.


aireilly (approved):
lgtm

aireilly (approved):
lgtm

PR #606, opened 2023-08-14
Do not report 'as well as per something'

jhradilek:
The current implementation of `RedHat.TermsErrors` causes `vale` to report the following error for sentences like "you can configure meta options per alert agent as well as per recipient":

```
 15:56   error       Use 'according to', 'as', or    RedHat.TermsErrors      
                     'as in' rather than 'as per'.
```

While "as well as" on its own is not recommended by the IBM Style Guide and gets reported by `RedHat.Usage`, the above error is out of place and the suggested wording is confusing.


aireilly (approved):
lgtm

PR #607, opened 2023-08-15
Update hard wrapped rule to accomodate sentence clauses

aireilly:
DL lists with short explanations that don't take full stops should not trigger the hard-wrapped rule.

Eg., these are not complete sentence clauses and should not hit the rule
 
```asciidoc
Hard drive:: Permanent storage for operating system and/or user files that are used on a daily basis
RAM:: Temporarily stores information that the CPU uses during the math operation
```


rohennes (approved):
/lgtm

PR #613, opened 2023-08-22
Detect problems in single-quote IDs

jhradilek:
In addition to double quotes, AsciiDoc allows the use of single quotes around ID values, for example:

```
[id='some-name']
```

I extended the regular expression to report unclosed single-quote ID values as well as detect mismatched quotes:

```
[id='some-name]
[id='some-name"]
[id="some-name']
```


aireilly (approved):
lgtm! thanks for the PR!

PR #616, opened 2023-08-25
Adding guidance for arch terms after SSG was updated

rohennes:
Fixes https://github.com/redhat-documentation/vale-at-red-hat/issues/199.

SSG was finalized with regard to terminology for different architiectures. Adding this back into Vale rules as a suggestion.


rohennes:
cc @max-cd @aireilly

aireilly (approved):
lgtm

PR #658, opened 2023-10-03
Changing the module context rule as it doesn't highlight anything in VS Code 

rohennes:
The rule to check for the presence of the :_content-type: in a module works but it doesn't show any line highlighting in VS Code. It only show the error on the cmd line. 
This PR changes the rule to a tengo script, which highlights the first line of the file if it doesn't find the variable.

### DIFF
--- a/.vale/styles/OpenShiftAsciiDoc/ModuleContainsContentType.yml
+++ b/.vale/styles/OpenShiftAsciiDoc/ModuleContainsContentType.yml
@@ -1,8 +1,29 @@
 ---
-extends: occurrence
-scope: raw
+extends: script
+message: "Module is missing the \":_mod-docs-content-type:\" variable."
 level: error
 link: https://github.com/openshift/openshift-docs/blob/main/contributing_to_docs/doc_guidelines.adoc#module-file-metadata
-message: "Module is missing the \"_mod-docs-content-type\" variable."
-min: 1
-token: '(?<!\/\/):_mod-docs-content-type:(\s)?(CONCEPT|PROCEDURE|REFERENCE|ASSEMBLY|SNIPPET)'
+scope: raw
+script: |
+  text := import("text")
+  matches := []
+
+  //trim extra whitespace
+  scope = text.trim_space(scope)
+  //add a newline, it might be missing
+  scope += "\n"
+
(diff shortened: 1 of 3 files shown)


.vale/styles/OpenShiftAsciiDoc/ModuleContainsContentType.yml
```diff
@@ -1,8 +1,29 @@
...
+script: |
+  text := import("text")
+  matches := []
+
+  //trim extra whitespace
+  scope = text.trim_space(scope)
+  //add a newline, it might be missing
+  scope += "\n"
+
+  context_regex := ":_content-type:(\\s)?(CONCEPT|SNIPPET|ASSEMBLY|PROCEDURE|REFERENCE)"
```
aireilly, line 16: Pretty sure the variable needs precisely one space after the variable name or AsciiDoctor complains

jhradilek:
```suggestion
  context_regex := "^:_(mod-docs-)?content-type:\\s+(CONCEPT|SNIPPET|ASSEMBLY|PROCEDURE|REFERENCE)"
```

The `modular-docs` project [recently changed](https://github.com/redhat-documentation/modular-docs/pull/209) the names of the attributes identifying the module type.

rohennes: Thanks @jhradilek - I don't think OpenShift have yet implemented this change, but when it does, I will need to update this rule

rohennes: > Pretty sure the variable needs precisely one space after the variable name or AsciiDoctor complains

Update. Does this work?

aireilly: IMO - if there is a failing on the VS Code plugin side, I would prefer to see that corrected rather than working around it with this update.

aireilly: Created https://github.com/ChrisChinchilla/vale-vscode/issues/14

PR #662, opened 2023-10-06
592: checking for correct assembly metadata - WIP

### LINKED ISSUE #592: Create Vale rules for all required OCP assembly metadata
Vale should check that each of the following are in place for assemblies. 

* If an attribute is found in the title, the rule should flag if the `_attributes/common-attributes.adoc` include is placed after the assembly title.
* :context: value should not include `_{context}` 

```
include::_attributes/common-attributes.adoc[]
:context: <unique-context-for-assembly>
= title
toc::[]
```

rohennes:
Fixes #592 
WIP - do not merge

### DIFF
--- a/.vale/styles/OpenShiftAsciiDoc/AssemblyContainsRequiredMetadata.yml
+++ b/.vale/styles/OpenShiftAsciiDoc/AssemblyContainsRequiredMetadata.yml
@@ -1,0 +1,37 @@
+---
+extends: script
+message: "Assembly is missing required metadata. Ensure you include metadata for attributes, TOC and context."
+level: error
+link: https://github.com/openshift/openshift-docs/blob/main/contributing_to_docs/doc_guidelines.adoc#assembly-file-metadata
+scope: raw
+script: |
+  text := import("text")
+  matches := []
+
+  //trim extra whitespace
+  scope = text.trim_space(scope)
+  //add a newline, it might be missing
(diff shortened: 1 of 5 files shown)


.vale/styles/OpenShiftAsciiDoc/AssemblyContainsRequiredMetadata.yml
```diff
@@ -0,0 +1,37 @@
+---
+extends: script
+message: "Assembly is missing required metadata. Ensure you include metadata for attributes, TOC and context."
```
aireilly, line 3:
```suggestion
message: "Assembly is missing required metadata and AsciiDoc toc elements. Ensure that you include common-attributes.adoc, toc::[], and :context: attributes."
```

aireilly (commented):
General comment. The rule as is implemented here catches the various errors but does not tell you which element is missing. This might be confusing for inexperienced users.

What about a series of rules, one for each element we want to police? 

then we could do something like this (for example):
```
extends: occurrence
message: 'Assembly should have a single "toc::[]" directive'
level: error
scope: raw
min: 1
max: 1
token: toc::\[\]
``` 

image

aireilly:
Planning to complete this? There has been no activity on this PR in months.

rohennes:
No don't have the bandwidth.

Issue #673, opened 2023-10-20
The word "reentrant" is reported as a warning


rolfedh:
**Describe the bug**
Vale doesn't seem to know the word "reentrant".

**To Reproduce**

1. Add the word "reentrant" to a text file.
2. Run vale against the file.

**Expected behavior**
"reentrant" should not show a warning

**Additional context**
IBM Style states that reentrant is one of several words that does not have a hyphen following the "re" prefix: "reredirect, reedit, reentrant, reexamine, resynchronize,"

https://www.ibm.com/docs/en/ztpf/2021?topic=entries-reentrant-programs

PR #674, opened 2023-10-20
Add "reentrant" to spelling rule

rolfedh:
Fixes https://github.com/redhat-documentation/vale-at-red-hat/issues/673


michelle-purcell (commented):
@rolfedh In the description of issue #673, the context confirms that "_re-entrant_" (with hyphen before re) is an invalid instance. 
As the style guide calls this out, might be an idea, to enhance the fix for issue #673.

michelle-purcell (approved):

Issue #128, opened 2021-12-13
RedHat.Admonitions: Keep admonitions to a minimum


themr0c:
Implement  https://redhat-documentation.github.io/supplementary-style-guide/#admonitions as a suggestion

message: Keep admonitions to a minimum
level: suggestion

when NOTE, IMPORTANT, WARNING, TIP is present

rolfedh:
I don't know a way to implement this in vale that is trully helpful: What this PR does is too simple and noisy.
 I would need to detect multiple admonitions stacked upon each other, or admonitions that contain multiple notes.

PR #677, opened 2023-10-20
Add rule for admonitions

rolfedh:
Fixes https://github.com/redhat-documentation/vale-at-red-hat/issues/128

### DIFF
--- a/.vale/styles/RedHat/Localization.yml
+++ b/.vale/styles/RedHat/Localization.yml
@@ -1,0 +1,12 @@
+---
+extends: existence
+ignorecase: false
+level: suggestion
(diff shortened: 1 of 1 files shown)


aireilly:
This will flag every instance of an admonition in every file in the repo. Consider using an occurrence rule: https://vale.sh/docs/topics/styles/#occurrence. 

even still, I wonder would this add a lot of noise in the rules output.

rolfedh:
I don't know a way to implement this in vale that is trully helpful: What this PR does is too simple and noisy.
I would need to detect multiple admonitions stacked upon each other, or admonitions that contain multiple notes.

PR #678, opened 2023-10-23
Enhance Red Hat spelling rule (YAML) to fix false positives on valid words

michelle-purcell:
This PR aims to remove the following false positive churn that we've seen recently on the Red Hat build of Quarkus doc project:

image

These are valid and commonly used US English words that are approved for use according to the latest Red Hat and IBM Style guidelines.


michelle-purcell:
@rohennes - FYI...  I changed `invokable` to `invocable` (in Merriam Webster) in the spelling rule and added a new Term Warning to detect `invokable` and suggest `invocable` instead.  I removed 'unsecure' as this might need more style discussion and investigation. Will handle in a later PR. Thanks :slightly_smiling_face:

rohennes (approved):
lgtm

Issue #679, opened 2023-10-23
Replace the :_content-type: variable with the :_mod-docs-content-type: variable


rolfedh:
**Is your feature request related to a problem? Describe.**

Replace the :_content-type: variable with the :_mod-docs-content-type: variable. 

For more information, see https://groups.google.com/a/redhat.com/g/content-services-list/c/DM-C3O_xXVQ 

**Additional context**

Note that the Vale at Red Hat style, RedHat, is reused by other repositories. Making this update here will propagate to other projects.

PR #680, opened 2023-10-23
Replace the :_content-type: variable with the :_mod-docs-content-type: variable

rolfedh:
Fixes https://github.com/redhat-documentation/vale-at-red-hat/issues/679
We won't merge this PR until we get the go-ahead to do so from @kalexand-rh.


rohennes:
This looks good @rolfedh but as mentioned on Slack, we should wait until OCP moves to the new variable naming convention. I believe that will happen in the next few weeks.

rolfedh:
@kalexand-rh would you review and approve, too?

rolfedh:
We won't merge this PR until we get the go-ahead to do so from @kalexand-rh.

kalexand-rh:
This looks good to me! I've updated the maintained OCP branches [here](https://github.com/openshift/openshift-docs/pull/67071), so this change should be good to incorporate.

aireilly (approved):
lgtm

PR #681, opened 2023-10-24
Add valid component names to headings and capitalization rules

michelle-purcell:
This PR aims to reduce false positive heading warnings and adds guidance for capitalization in Red Hat style rules for the following names, which are commonly used in Red Hat Runtimes and developer products:

- Liquibase
- Mockito
- OpenTelemetry (OTel)
- OTel
- Temurin

### DIFF
--- a/.vale/fixtures/RedHat/CaseSensitiveTerms/testvalid.adoc
+++ b/.vale/fixtures/RedHat/CaseSensitiveTerms/testvalid.adoc
@@ -164,6 +164,7 @@ LDAPS
 librados
 librbd
 Linux
+Liquibase
 LUN
 management CLI
 management console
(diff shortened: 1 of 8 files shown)


rohennes:
Looks great, just one thing, does OTel get used on it's own after the first definition? The PascalCamelCase rule was flagging the acronym Otel on it's own. Perhaps you could add an exception if OTel is used on its own?

image

michelle-purcell:
@rohennes - https://github.com/redhat-documentation/vale-at-red-hat/pull/681#issuecomment-1781036043

:+1:  Thanks so much for your review. I added an exception for `OTel` to the PascalCamelCase rule and fixture.

rohennes (approved):
lgtm

PR #682, opened 2023-10-24
Fix heading and case-sensitive rule for role-based access control (RBAC)

michelle-purcell:
This PR makes a small correction to the style guidance rule for _role-based access control (RBAC)_.

This security term should be all lowercase (no capital _R_ in Role), unless at the start of a sentence, which is already detected and handled by vale-at-red-hat rules.

### DIFF
--- a/.vale/fixtures/RedHat/CaseSensitiveTerms/testvalid.adoc
+++ b/.vale/fixtures/RedHat/CaseSensitiveTerms/testvalid.adoc
@@ -253,7 +253,7 @@ Red Hat Virtualization Manager
 RedBoot
 Redis
 RESTEasy
-Role-based access control
+role-based access control
 Rolling Application Stream
 Rolling Application Streams
 Rolling Stream
--- a/.vale/styles/RedHat/CaseSensitiveTerms.yml
+++ b/.vale/styles/RedHat/CaseSensitiveTerms.yml
@@ -251,7 +251,7 @@ swap:
   RHDS: Red Hat Directory Server
   RHEL host|RHEL-H: Red Hat Enterprise Linux host
   RHV: Red Hat Virtualization
-  Role Based Access Control|Role-Based Access Control|role based access control: Role-based access control
+  Role Based Access Control|Role-Based Access Control|role based access control: role-based access control
   rolling application stream: Rolling Application Stream
   rolling application streams: Rolling Application Streams
   rolling stream: Rolling Stream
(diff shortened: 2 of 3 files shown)


.vale/fixtures/RedHat/CaseSensitiveTerms/testvalid.adoc
```diff
@@ -253,7 +253,7 @@ Red Hat Virtualization Manager
 RedBoot
 Redis
 RESTEasy
-Role-based access control
+role-based access control (RBAC)
```
rohennes, line 256: Perhaps leave out the acronym? It may not be required in all instances in the docs, for example if you use the term infrequently in a module? There's also a separate rule for acronyms I believe

.vale/styles/RedHat/CaseSensitiveTerms.yml
```diff
@@ -251,7 +251,7 @@ swap:
   RHDS: Red Hat Directory Server
   RHEL host|RHEL-H: Red Hat Enterprise Linux host
   RHV: Red Hat Virtualization
-  Role Based Access Control|Role-Based Access Control|role based access control: Role-based access control
+  Role Based Access Control|Role-Based Access Control|role based access control: role-based access control (RBAC)
```
rohennes, line 254: Same comment regarding the acronym

rohennes:
> This PR makes a small correction to the style guidance rule for _role-based access control (RBAC)_.
> 
> This security term should be all lowercase (no capital _R_ in Role), unless at the start of a sentence, which is already detected and handled by vale-at-red-hat rules.

Just a question as it's something I encoutered in another PR. Do we have a rule that checks for an initial at the start of a sentence? I don't think we do actually.

.vale/fixtures/RedHat/CaseSensitiveTerms/testvalid.adoc
michelle-purcell: @rohennes - Good point :+1: Fixed. Thanks.

michelle-purcell:
@rohennes - https://github.com/redhat-documentation/vale-at-red-hat/pull/682#issuecomment-1780919641

I'm not aware of one. I fixed the items you mentioned and tested the associated initialisms in various content elements, like headings and paragraphs and I didn't see any errors so far.

Thanks so much for your review :tada:

rohennes (approved):
lgtm

PR #683, opened 2023-10-26
Update installation instructions

### LINKED ISSUE #87: Update installation instructions
> Souvik Sarkar  1 hour ago: If you place the `.ini` files within the project directory, then you might have to either add them as exceptions in `.gitignore` (needs buy in from the greater community of contributors), or  manually stage the modified files every time you make changes and commit, instead of the usual `git add .`

This item is a more conceptual and a bit of an edge case: the current instructions for installing the Vale at Red Hat style bypass this issue altogether. However, we should probably discuss this as an alternative configuration scenario.

rolfedh:
Fixes #87


aireilly (approved):
lgtm

PR #684, opened 2023-10-31
Update recommended vale extension

r0ckarong:
The currently mentioned extension is deprecated and VS Code will suggest to use this newer extension.


aireilly (approved):
lgtm

PR #690, opened 2023-11-09
Added rule: for instance

### LINKED ISSUE #687: Add "for instance" to the rules
**Is your feature request related to a problem? Describe.**
IBM Style clearly discourages "for instance" and encourages using "for example" instead.
https://www.ibm.com/docs/en/ibm-style?topic=word-usage#do-not-use-for-instance-adverb

**Describe the solution you'd propose**
Add "for instance" to the failing rules. However, there are valid occurrences of this phrase, for example "Configuring the Compute Service for Instance Creation", "Support for instance domain change-deltas", and "for instances where...".

Therefore, adding a space before and a comma after the expression should limit the rate of false positives.

**Additional context**
This is quite a common expression and without the check it occurs frequently in existing RH docs.

gaurav-nelson:
Fixes #687 

New rule: Checks if there is a space before and a comma following "for instance". 

**question:** ❔ Should we also check for "for instance" with spaces before and after?

### DIFF
--- a/.vale/fixtures/RedHat/TermsErrors/testinvalid.adoc
+++ b/.vale/fixtures/RedHat/TermsErrors/testinvalid.adoc
@@ -470,3 +470,4 @@ zero out
 zeroes
 zonegroup
 Zulu time
+There are many ways to write, for instance, using a pencil.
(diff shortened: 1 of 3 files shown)


aireilly (commented):
regex lgtm, but wouldn't it be better to include this fix in the `TermsErrors.yml` rule? then we would be able to include a quickfix with the Substitution rule.
image

`'(?<= )for instance(?=,)': for example`

aireilly (approved):
lgtm

PR #697, opened 2023-11-29
Update with the URL of the new supported Vale VSCode extension

rolfedh:
I replaced a link to the deprecated vale-server extension with one for the newer _Vale VSCode_ extension.
Another topic, modules/user-guide/pages/using-vale-in-the-ide.adoc already links to the Vale VSCode extension.


aireilly (approved):
Lgtm

Issue #698, opened 2023-12-08
Create a number ranges rule


aireilly:
![image](https://github.com/redhat-documentation/vale-at-red-hat/assets/74046732/bc0d383e-d394-4adb-8354-8a09a56fcc75)

https://www.ibm.com/docs/en/ibm-style?topic=measurement-numbers#ranges

PR #699, opened 2023-12-12
Adding En dash rule

aireilly:
Fixes https://github.com/redhat-documentation/vale-at-red-hat/issues/698

> Generally, use an en dash (–) without spaces on each side to indicate a range of numbers. If you list numbers along with other characters, unless the context determines a different order of the characters, list numeric characters first, followed by alphabetic characters (uppercase and then lowercase), followed by other characters in ASCII order.


modules/reference-guide/pages/endashes.adoc
```diff
@@ -0,0 +1,18 @@
+:navtitle: Endashes
+:keywords: reference, rule, endashes
+
+= En Dashes
```
apinnick, line 4: Small nit. I have not checked the other "pages". Do they use title case or sentence case?

apinnick (approved):
Apart from a query about the "page" title, LGTM.

modules/reference-guide/pages/endashes.adoc
aireilly: Cool. Sentence case I guess. thanks!

aireilly:
In retrospect, I am going to close this un-merged. Terms like the below perfectly valid and will get caught in this rule.

```
ip-10-0-0-1.us-east-2.compute.internal
775898c3-ecc2-41a4-b98b-a4cd5ae56fd0
2023-09-22
```

Issue #672, opened 2023-10-20
Use double quotes, not single, in IDs


rohennes:
I don't think it causes a build issue, but this is non-standard:

[id='support-overview-remote-health-monitoring']

Should use double quotes:

[id="support-overview-remote-health-monitoring"]

michellemacrh:
Interesting to note, thanks for the feedback 👍 

"As a project style rule" - Does that refer to a .vale.ini file outside of the vale-at-red-hat repo?

rohennes:
I think this came from a peer review comment in the OpenShift docs. As it's not bad AsciiDoc, perhaps it's better not to flag as an error in the AsciiDoc rule set. 

As Aidan alluded to, if this went against style guidelines for a particular project, for example RHEL or OpenShift or whoever, you could create a rule for it and include it in the project's rule set. And yes, the .vale.ini file is how you configure Vale in a repo, for example, what rules you want turned on or off in a rule set, or what rule sets (or packages) you want to use etc.

michellemacrh:
Thanks for your comments @rohennes. I'm still getting to grips with using Vale so this info is helpful :) 

In terms of helping to resolve this issue, is it a matter of moving this rule into OpenShiftAsciiDoc, or does this need to be addressed within the OpenShift docs themselves?

rohennes:
Typically these types of rules might hang off some project-specific guidance. We don't have a rule about this in the OpenShift style guidelines though. So it's probably solving an issue that doesn't really exist. I got a review comment about it but perhaps that was just one reviewer's preference rather then a hard rule.

Sorry if this issue was a wild goose chase but probably need to drop it unless you can find some project that really disallows this. The updates in your PR looked good otherwise!

rohennes:
You could try asking the mod docs people if they want to mandate for double-quotes by opening an issue? You could add it to the Red Hat rule set then. It's a bit of a long shot though. https://github.com/redhat-documentation/modular-docs/issues

michellemacrh:
Thanks @rohennes, I'll take a look :)

aireilly:
We support single and double quotes in the rules FYI: https://github.com/redhat-documentation/vale-at-red-hat/pull/613

I don't think we need to change this tbh

PR #711, opened 2024-01-29
vale-issue-672 - Use double quotes not single quotes in IDs

michellemacrh:
This PR addresses #672

My regex knowledge leaves a lot to be disired but I made an attempt at adding this rule.

As it relates to AsciiDoc syntax, I added to the AsciiDoc package.

The tests pass when I run `vale .` but this could do with a sanity check.

Any and all feedback welcome :)


aireilly:
Again - thanks for the PR. We appreciate it! Unfortunately, we won't merge this since single and double quotes are both perfectly valid in AsciiDoc once they are not mixed. 

Reach out on slack if you'd like to be assigned some other tasks! There is lots in the queue :)

michellemacrh:
No problem, thanks for the feedback, I'll close this out :) I certainly will reach out on Slack, thank you!

PR #714, opened 2024-02-09
read/write should not trigger the slash rule + adjust scope for Slash.yml

aireilly:


rohennes (approved):
Do the additions to the Slash rule need to be added in the testvalid.adoc fixture? Otherwise looks good to me

PR #717, opened 2024-02-09
Replacing double quotes in alert messages with single quotes.

rohennes:
Double quotes within the message body causes problems for further processing with other programs. For example, I get confused JSON converting to JSON
 
 `{"body":"Module is missing the "_mod-docs-content-type" variable.","path":"test.adoc","line":1}`
 
 Changing the 3 instances in the OCP rule set so it will read like the following, which is easier JSON to process.
 
` {"body":"Module is missing the '_mod-docs-content-type' variable.","path":"test.adoc","line":1}`
 
 Using single quotes in messages is also the format used in other rule sets.


aireilly (approved):
lgtm

PR #720, opened 2024-02-27
Tightens what we fail as errors in RH Vale Styles

aireilly:
1. Tightens what we fail as errors in RH Vale Styles
2. Adds quotes to errors messages where they are missing


rohennes (approved):
/lgtm

PR #729, opened 2024-03-06
Adding unset attribute rule

aireilly:
Adds a new AsciiDoc rule that highlights when an attribute is not unset in the same file. 

The rule fires for an attribute that is not unset in the file.

### DIFF
--- a/.vale/fixtures/AsciiDoc/UnsetAttributes/.vale.ini
+++ b/.vale/fixtures/AsciiDoc/UnsetAttributes/.vale.ini
@@ -1,0 +1,5 @@
+; Vale configuration file to test the `UnsetAttributes` rule
+StylesPath = ../../../styles
+MinAlertLevel = error
+[*.adoc]
+AsciiDoc.UnsetAttributes = YES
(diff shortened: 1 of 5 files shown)


aireilly:
This rule does not account for specific ifeval situations. 

```
ifeval::["{context}" == "build-strategies-docker"]
:dockerstrategy:
endif::[]

<module text>

ifeval::["{context}" == "build-strategies-docker"]
:!dockerstrategy:
endif::[]
```

Update the rule to respect ifeval boundaries

aireilly:
> This rule does not account for specific ifeval situations.
> 
> ```
> ifeval::["{context}" == "build-strategies-docker"]
> :dockerstrategy:
> endif::[]
> 
> <module text>
> 
> ifeval::["{context}" == "build-strategies-docker"]
> :!dockerstrategy:
> endif::[]
> ```
> 
> Update the rule to respect ifeval boundaries. For example, how can we validate this? https://github.com/aireilly/openshift-docs/blob/add-symlinks-to-codeowners/modules/cco-ccoctl-configuring.adoc?plain=1

Hacked around this by just counting set and unset attributes

aireilly:
> /cc

What does this mean?

aireilly:
@jeana-redhat Given upcoming tooling changes, I don't think I will be able to spend more time getting this into a better state for AsciiDoc, this will have to do.

jeana-redhat:
> > /cc
> 
> What does this mean?

well, in other repos, it seems to mark you as a reviewer. I was planning to come back around on this :sweat_smile:

aireilly:
Merging this since it's reasonably useful and we're moving off AsciiDoc so why not? :shrug:

PR #732, opened 2024-03-06
Clean out multi-line AsciiDoc comments in raw script rules

aireilly:
Since script rules operate on raw AsciiDoc (not HTML), the rules need to handle commented content explicitly.

This update cleans out the comments before linting the source file.

```
// clean out multi-line comments
scope = text.re_replace("(?s) *(\n////.*?////\n)", scope, "")
```


aireilly:
To test, use fmt.println to see the scope after the comment is removed

rohennes (approved):
/lgtm - thanks Aidan

PR #734, opened 2024-03-08
Editing alert message as the line number of the alert may not be where the problem is

rohennes:
Making it clearer that the user should search the entire file unbalanced if statements, not just the line where the alert is triggering.


aireilly (approved):
/lgtm

PR #742, opened 2024-03-20
Add utilizing to SimpleWords.yml

michellemacrh:
Add utilizing to SimpleWords.yml

"utilize" and "utilization" are included in SimpleWords.yml but I think "utilizing" should also be added. 

Discussed in Vale Slack channel and I agreed to work on a PR.


aireilly (approved):
/lgtm

PR #745, opened 2024-03-21
Allow for VT-d, the product is Intel VT-d so should be allowed

rohennes:
Fixes incorrect flagging found in ocpdocs-vale-bot: https://github.com/openshift/openshift-docs/pull/72761#discussion_r1532155937


aireilly (approved):
/lgtm

PR #747, opened 2024-03-22
Fix spacing rule 

aireilly:
In the CI, the spacing rule makes no sense. This update expands the regex to return a meaningful message.
image


rohennes (approved):
/lgtm 

Thanks!

PR #748, opened 2024-03-25
Kubelet Stats Receiver is a new feature and kubelet can use an upper case K in this instance

rohennes:
https://github.com/openshift/openshift-docs/pull/73626/files#r1535802857 - example false positive. I don't understand why the error message doesn't suggest replacing with "kubelet" - I can't replicate this locally - but at least this rule won't fire on this product name


aireilly:
Is this something to do with it being in a heading.

aireilly (approved):
/lgtm

PR #750, opened 2024-03-27
Removing Azure as an error term 

rohennes:
The SSG allows Azure as a term if it's already been defined previously as Microsoft Azure. So perhaps a more complex rule is required instead. 

Also too many errors on Azure product names.


aireilly (approved):
/lgtm

PR #763, opened 2024-04-16
Excluding term that is a valid OCP operator

rohennes:
Adding an exception for the term Kernel Module Management as this is a valid OCP operator. 

See false positive here: https://github.com/openshift/openshift-docs/pull/74713#discussion_r1567479450


aireilly (approved):
/lgtm

PR #780, opened 2024-04-29
Fix RedHat DoNotUseTerms validation issues for Red{nbsp}Hat

gaurav-nelson:
Updated the regex to not match `Red{nbsp}Hat`.

Currently it shows a suggestion:
```
Do not use colors to describe something unless it is also described non-visually, 
for example, with the name of the item.
```

### DIFF
--- a/.vale/styles/RedHat/DoNotUseTerms.yml
+++ b/.vale/styles/RedHat/DoNotUseTerms.yml
@@ -4,6 +4,7 @@ ignorecase: true
 level: warning
 link: https://redhat-documentation.github.io/vale-at-red-hat/docs/main/reference-guide/donotuse/
 message: "%s"
+scope: raw
 swap:
   # Start each error message with "Do not use ..."
   # Error messages must be single quoted.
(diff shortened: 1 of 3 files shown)


aireilly:
Need to merge this one. Can you fix the tests and we'll get it in.

.vale/styles/RedHat/DoNotUseTerms.yml
```diff
@@ -4,6 +4,7 @@ ignorecase: true
 level: warning
 link: https://redhat-documentation.github.io/vale-at-red-hat/docs/main/reference-guide/donotuse/
 message: "%s"
+scope: raw
```
gaurav-nelson, line 7: @aireilly I don't know why but this rule wouldn't work without `scope: raw` 
At https://github.com/errata-ai/vale/issues/233#issuecomment-693607063 it says that regex is independent of the `raw` scope. ¯\_(ツ)_/¯

aireilly: I'll have a look.

aireilly: Ah. Is it because Vale works on the HTML, and when we generate HTML we get 
```
![image](https://github.com/redhat-documentation/vale-at-red-hat/assets/74046732/139045b5-9341-4658-819a-072e25f7855d)
```

aireilly: I would be inclined to remove the entire "Red" element from the rule if we can't make it work without scope.

Either that, or remove `ignorecase: true` and `scope: raw` and go with:

```
  \bred\b(?!-? *hat)|orange|yellow|green|blue|indigo|violet: 'Do not use colors to describe something unless it is also described non-visually, for example, with the name of the item.'
```

aireilly: :shrug:

aireilly: Setting "raw" also means that we would start linting for all these words in all text in an AsciiDoc file, incl inside codeblocks, and other mark up. Not ideal.

gaurav-nelson:
Thank you for reviewing @aireilly Your comments make sense. Another thing we can try is to add `Red{nbsp}Hat` in the dictionary instead of this. I'll try it locally.

aireilly:
Felt this was important enough to complete so did it here: https://github.com/redhat-documentation/vale-at-red-hat/pull/780

Thanks for raising the PR

PR #788, opened 2024-05-15
Creating a new localization rule set for vale

reachlekha:
Creates a new localization rule set for highlighting common words that can cause trouble for translators. 
The first cut includes the following word categories :
- Modals
- Commonly misused words
- Phrasal words
- Words with multiple meanings
 
 @chloeredhat @aireilly @rolfedh @rohennes - Please help review! Thanks!


.vale/styles/Localization/CommonlyMisusedWords.yml
```diff
@@ -0,0 +1,10 @@
+---
+extends: existence
+ignorecase: true
+level: warning
+link: https://redhat-documentation.github.io/vale-at-red-hat/docs/main/reference-guide/termswarnings/
+message: "Avoid terms that are ambiguous and problematic in a localization context"
```
aireilly, line 6: Can we add some specific detail for each of these rules instead of a generic message for all.

chloeredhat: @reachlekha I just noticed that this message is incorrect, sorry I missed that. It's not that these words are wrong, so you should remove the language of "avoid". 

The purpose of this section is making sure users are using the proper term. For example, many people write "acronym" when they really mean abbreviation or initialism. Likewise, it's not that people can never use "allow". But, people frequently confuse it with the word "enable" (allow = permission, enable = ability)

.vale/fixtures/Localization/Modals/testinvalid.adoc
```diff
@@ -0,0 +1,13 @@
...
+would
+will
+supposed to
+ought to be
+as
+at once
+once
+while
+since
+allow
```
chloeredhat, line 12: Is there a reason words that aren't modals are in this? These words are not modals:

* as
* at once
* while
* since
* allow

.vale/fixtures/Localization/Modals/testvalid.adoc
```diff
@@ -0,0 +1,5 @@
+might
+enable
```
chloeredhat, line 2: Enable is not a modal

.vale/styles/Localization/Modals.yml
```diff
@@ -0,0 +1,15 @@
+---
+extends: existence
+ignorecase: true
+level: warning
+link: https://redhat-documentation.github.io/vale-at-red-hat/docs/main/reference-guide/termswarnings/
+message: "Avoid terms that are ambiguous and problematic in a localization context"
```
chloeredhat, line 6: Maybe make more specific to modals: "Avoid using modals because they can introduce uncertainty"

.vale/styles/Localization/WordswithMultipleMeanings.yml
```diff
@@ -0,0 +1,14 @@
+---
+extends: existence
+ignorecase: true
+level: warning
+link: https://redhat-documentation.github.io/vale-at-red-hat/docs/main/reference-guide/termswarnings/
+message: "Avoid terms that are ambiguous and problematic in a localization context"
```
chloeredhat, line 6: "Avoid words that have multiple meanings. For example, instead of words like "as" or "since" which have multiple meanings, instead use "because".

.vale/styles/Localization/CommonlyMisusedWords.yml
reachlekha: > Can we add some specific detail for each of these rules instead of a generic message for all.

Sure!

.vale/fixtures/Localization/Modals/testvalid.adoc
reachlekha: Removed it from the list. Thanks!

.vale/styles/Localization/Modals.yml
reachlekha: Updated as per suggestion! Thanks!

.vale/styles/Localization/WordswithMultipleMeanings.yml
reachlekha: Updated as per suggestion! Thanks!

.vale/styles/Localization/CommonlyMisusedWords.yml
reachlekha: Updated as per suggestion! Thanks!

.vale/styles/Localization/Modals.yml
```diff
@@ -0,0 +1,15 @@
+---
+extends: existence
+ignorecase: true
+level: warning
+link: https://redhat-documentation.github.io/vale-at-red-hat/docs/main/reference-guide/termswarnings/
+message: "Avoid using modals because they can introduce uncertainty"
```
chloeredhat, line 6: Suggested edit:

"Avoid using modals like 'should', 'may', or 'will' because they can introduce uncertainty."

reachlekha: Thanks! Updated as suggested!

aireilly:
Have you reviewed https://www.ibm.com/docs/en/ibm-style?topic=medium-global-audiences? I bet there are lots of things we could incorporate

.vale/fixtures/Localization/WordswithMultipleMeanings/testvalid.adoc
```diff
@@ -0,0 +1,7 @@
+after
```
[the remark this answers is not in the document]
reachlekha: Updated as suggested.

.vale/styles/Localization/CommonlyMisusedWords.yml
```diff
@@ -0,0 +1,10 @@
+---
+extends: existence
+ignorecase: true
+level: warning
+link: https://redhat-documentation.github.io/vale-at-red-hat/docs/main/reference-guide/termswarnings/
+message: "Suggest to use proper terms to reduce misconceptions. For example- Do not use allow with the meaning of enable. (allow = permission, enable = ability) "
```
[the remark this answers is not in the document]
reachlekha: Updated as suggested. Thanks!

.vale/styles/RedHat/PassiveVoice.yml
```diff
@@ -64,11 +64,15 @@ tokens:
   - fought
   - found
   - frozen
+  - get
```
chloeredhat, line 67: so these need to be part of the regex, not on their own. If we put them here on their own they will flag way too many things. The regex is    \b(am|are|were|being|is|been|was|be)\b\s*

.vale/fixtures/Localization/WordswithMultipleMeanings/testinvalid.adoc
```diff
@@ -2,4 +2,7 @@ as
 at once
 once
 while
-since
\ No newline at end of file
+since
+as soon as
```
chloeredhat, line 6: The as soon as onwards should actually be in the Valid file. We want Vale to catch "as" on its own, but not expressions like "as much as" or "as soon as"

.vale/styles/Localization/WordswithMultipleMeanings.yml
```diff
@@ -3,10 +3,13 @@ extends: existence
 ignorecase: true
 level: warning
 link: https://redhat-documentation.github.io/vale-at-red-hat/docs/main/reference-guide/termswarnings/
-message: "Avoid words that have multiple meanings. For example, instead of words like "as" or "since" which have multiple meanings, use "because""
+message: "Avoid words that have multiple meanings. For example, instead of words like "as" or "since" which have multiple meanings, use "because"."
 # source: https://github.com/redhat-documentation/vale-at-red-hat/issues/515
 tokens:
   - as
+  - as soon as
```
chloeredhat, line 10: Are these here as invalid? See comment above about these being valid

.vale/fixtures/RedHat/PassiveVoice/testinvalid.adoc
```diff
@@ -54,6 +54,8 @@ It was forsaken
 It was fought
 It was found
 It was frozen
+I may get late
```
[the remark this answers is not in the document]
reachlekha: Done! Updated as suggested!

.vale/styles/Localization/WordswithMultipleMeanings.yml
reachlekha: Done! Updated as suggested!

.vale/styles/RedHat/PassiveVoice.yml
reachlekha: Done! Updated as suggested!

chloeredhat (approved):

aireilly (commented):
Some good phrasal verb terms here pulled from the ISG: https://github.com/redhat-documentation/vale-at-red-hat/pull/809

aireilly:
> Some good phrasal verb terms here pulled from the ISG: #809

Also maybe, "picks? up," "gives? up," or "sets? up"

.vale/styles/Localization/CommonlyMisusedWords.yml
```diff
@@ -0,0 +1,12 @@
+---
+extends: substitution
+ignorecase: true
+level: warning
+link: https://redhat-documentation.github.io/vale-at-red-hat/docs/main/reference-guide/termswarnings/
+message: "%s"
+swap:
+  # Start each error message with "Do not use ..."
+  # Error messages must be single quoted.
+allow: 'Only use allow if you are talking about permission. For ability, use enable.'
```
aireilly, line 10: Also not certain about this one. If the goal of the rule set is to highlight localization errors, and there are very definite instances where the word is correctly used, I would tend to leave it out of the rule set, unless you can expand on the rule and highlight grammatical usages where the word definitely is misused.

.vale/styles/Localization/CommonlyMisusedWords.yml
```diff
@@ -0,0 +1,12 @@
...
+acronym: 'Acronyms are only for abbreviations that are pronounced as a word, like RHEL. Use initialism for abbreviations that are spelled out. Alternatively, use abbreviation for either situation.'
```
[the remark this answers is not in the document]
chloeredhat: Fair point, we can delete since it is not frequently used in writing!

.vale/styles/Localization/PhrasalVerbs.yml
```diff
@@ -0,0 +1,18 @@
+---
+extends: substitution
```
chloeredhat, line 2: Per Aidan's suggestion, we can add these that Gaurav thought of: @Chandralekha-RedHat 

call up
carry on
carry out
click on
find out
print out
start up
switch on
tap on

chloeredhat: Aidan also suggested:
"Also maybe, "picks? up," "gives? up," or "sets? up""

I am fine with 
picks up
gives up

But I do NOT think we should add "sets up" because set-up is a valid term.

.vale/styles/RedHat/SelfReferentialText.yml
```diff
@@ -12,3 +12,11 @@ tokens:
...
   - this section
   - this subsection
+  - The following
+  - The previous
+  - This is
+  - These are
+  - Those are
+  - Theirs
+  - Hers
+  - His
```
chloeredhat, line 22: @reachlekha I looked through the that vs. which content in Mie's document. I am not sure how common what she is describing is, so we might not need to include it. However, it would involve putting this in testinvalid:

",which"

Basically, you shouldn't have a comma in front of which. But, I think this is more of a grammar thing than localization. We don't have rules about excessie comma use anywhere else in Vale.

.vale/styles/Localization/CommonlyMisusedWords.yml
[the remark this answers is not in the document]
reachlekha: Done! Removed the rule for "acronym".

.vale/styles/Localization/PhrasalVerbs.yml
reachlekha: Thanks! Updated as suggested!

.vale/styles/RedHat/SelfReferentialText.yml
reachlekha: Thanks! Updated as suggested!

.vale/styles/Localization/CommonlyMisusedWords.yml
Chandralekha-RedHat: @chloeredhat Thanks! Updated as suggested!

aireilly: The test cases look good, but why aren't these test cases reflected in the actual rule? Also, @chloeredhat's suggestions have the word "allows", your rule is just for the word "allow".

aireilly: None of the terms in .vale/fixtures/Localization/CommonlyMisusedWords/testinvalid.adoc will be flagged by this rule.

Chandralekha-RedHat: Good catch! Thanks for that! I updated the testvalid and testinvalid files but not the yaml file. Here I go update it!

.vale/styles/Localization/PhrasalVerbs.yml
```diff
@@ -0,0 +1,24 @@
...
+action:
+  name: replace
+swap:
+  # Start each error message with "Do not use ..."
+  # Error messages must be single quoted.
+do: carry out
+allow | enable: allow for
+continue: carry on
+depend: count on
+result: end up
```
aireilly, line 16: This will flag every instance of result in source. There will be many false positives.

.vale/styles/Localization/PhrasalVerbs.yml
```diff
@@ -0,0 +1,24 @@
...
+imply: get at
+click: click on
```
aireilly, line 18: This will flag every instance of click, even if it is followed by "on".

.vale/styles/Localization/PhrasalVerbs.yml
```diff
@@ -0,0 +1,24 @@
+---
```
aireilly, line 1: General comment for this rule - as it is laid out below, there will be a huge amount of false positives for this rule. Need to reduce the false positives to make this useful.

.vale/styles/RedHat/SelfReferentialText.yml
```diff
@@ -12,3 +12,11 @@ tokens:
   - this chapter
   - this section
   - this subsection
+  - The following
```
aireilly, line 15: https://www.ibm.com/docs/en/ibm-style?topic=word-usage#following

Can we make this more useful by creating a sequence rule that returns instances of "following" where it is not in the adjectival position before a noun? 

https://github.com/aireilly/vale-at-red-hat/blob/add-pos-using-rule/.vale/styles/RedHat/Using.yml

https://vale.sh/docs/topics/styles/#sequence

Hint:

.vale/styles/Localization/PhrasalVerbs.yml
```diff
@@ -0,0 +1,24 @@
...
+ignorecase: true
+level: warning
+link: https://redhat-documentation.github.io/vale-at-red-hat/docs/main/reference-guide/termswarnings/
+message: "Avoid using phrasal verbs. Consider using '%s' rather than '%s'."
+action:
+  name: replace
+swap:
+  # Start each error message with "Do not use ..."
+  # Error messages must be single quoted.
+do: carry out
```
aireilly, line 12: I think all these rules are mixed up. They should be in the form `bad: good`

.vale/fixtures/Localization/AdjectivesWithNouns/.vale.ini
```diff
@@ -0,0 +1,5 @@
+; Vale configuration file to test the `CommonlyMisusedWords` rule
```
ccoVeille, line 1: I'm unsure about this comment, it looks like a copy pasta

.vale/fixtures/Localization/PhrasalVerbs/.vale.ini
```diff
@@ -0,0 +1,5 @@
+; Vale configuration file to test the `CommonlyMisusedWords` rule
```
ccoVeille, line 1: Same

.vale/styles/Localization/PhrasalVerbs.yml
```diff
@@ -0,0 +1,25 @@
...
+ignorecase: true
+level: warning
+link: https://redhat-documentation.github.io/vale-at-red-hat/docs/main/reference-guide/termswarnings/
+message: "Avoid using phrasal verbs. Consider using '%s' rather than '%s'."
+scope: sentence
+action:
+  name: replace
+swap:
+  # Start each error message with "Do not use ..."
+  # Error messages must be single quoted.
```
ccoVeille, line 12: I don't get this comment.

It seems related to the error message,not the swap section

.vale/styles/Localization/CommonlyMisusedWords.yml
```diff
@@ -0,0 +1,17 @@
+---
+extends: substitution
+ignorecase: true
+level: warning
+link: https://redhat-documentation.github.io/vale-at-red-hat/docs/main/reference-guide/termswarnings/
+message: 'Only use allow if you are talking about permission. For ability, use enable.'
```
ccoVeille, line 6: I know it's about termswarning, but there is something strange to me.

Either I'm missing something about Vale, either this token list will report everything that contains it.

I'm worried about the fact you will report a warning for something that could be OK in the sentence (something about permission)

Shouldn't you use a `level: suggestion` for this one?

.vale/styles/Localization/AdjectivesWithNouns.yml
```diff
@@ -0,0 +1,11 @@
+---
+extends: sequence
+message: "Always use 'following' as an adjective before a noun for clarity and grammatical correctness. For example, use 'Complete the following steps' instead of 'Complete the following'."
```
chloeredhat, line 3: This is correct but I think it would be more concise to say:
"Do not use adjectives without a noun following them. For example, use 'Complete the following steps' instead of 'Complete the following'."

.vale/styles/RedHat/SelfReferentialText.yml
```diff
@@ -12,11 +12,3 @@ tokens:
   - this chapter
   - this section
   - this subsection
-  - The following
-  - The previous
-  - This is
```
chloeredhat: Can we add these to AdjectivesWithNouns? These are other examples (everything from 'this is' to 'those are' but not the last 3, don't include those)

Some "correct" versions would be things like "This is" --> "This function is" 
or 
"Those are" --> "Those instructions are"

.vale/fixtures/Localization/AdjectivesWithNouns/.vale.ini
Chandralekha-RedHat: Good catch! Updated! Thanks!

.vale/styles/Localization/AdjectivesWithNouns.yml
Chandralekha-RedHat: Done! Updated as suggested!

.vale/styles/Localization/CommonlyMisusedWords.yml
Chandralekha-RedHat: Done! Updated as suggested!

.vale/styles/RedHat/SelfReferentialText.yml
Chandralekha-RedHat: Done! Updated as suggested!

.vale/styles/Localization/PhrasalVerbs.yml
Chandralekha-RedHat: @aireilly Could you please help clarify this rule?

.vale/fixtures/Localization/PhrasalVerbs/.vale.ini
```diff
@@ -0,0 +1,5 @@
+; Vale configuration file to test the `PhrasalVerbs` rule
+StylesPath = ../../../styles
+MinAlertLevel = warning
+[*.adoc]
+Localization.TermsWarnings = YES
```
aireilly, line 5:
```suggestion
Localization.PhrasalVerbs = YES
```

.vale/fixtures/Localization/WordswithMultipleMeanings/.vale.ini
```diff
@@ -0,0 +1,5 @@
+; Vale configuration file to test the `WordswithMultipleMeanings` rule
+StylesPath = ../../../styles
+MinAlertLevel = warning
+[*.adoc]
+Localization.TermsWarnings = YES
```
aireilly, line 5:
```suggestion
Localization.WordswithMultipleMeanings = YES
```

aireilly:
Can we close this? There has been no activity on this PR in months.

apinnick:
Closing because no activity

PR #799, opened 2024-06-07
Add part-of-speech rule for [noun + using]

aireilly:
The existing `TermsSuggestions.yml` rule for `"(?<![Bb]y) using": by using|that uses` fires lots of false positives for "using". For example, "You are using the computer." should not raise a Vale error.

https://studio.vale.sh/s/0d3e250c6fcaed40fb64cb8b797f7f7f

Tested against openshift-docs with:

```cmd
$ vale ./modules --filter='.Name == "RedHat.Using"' --glob='*.adoc' --output=line
```

PR also removes `.github/workflows/web-ide-link.yml` and updates the validate-rules job to use the latest version of Vale, which is required for this rule.

Also various clean up items for new vale version 3.7.0.


rohennes (approved):
lgtm - just minor comments

.vale/styles/RedHat/Using.yml
```diff
@@ -0,0 +1,14 @@
+---
+extends: sequence
+message: "Use 'by using' instead of 'using' when it follows a noun for clarity and grammatical correctness."
+link: https://redhat-documentation.github.io/vale-at-red-hat/docs/main/reference-guide/using/
+level: warning
+action:
+  name: edit
+  params:
+    - regex
```
rohennes, line 9: Just checking the regex param is required?

aireilly: It does look funny but yes it needs to be there. At least in my testing. https://vale.sh/docs/topics/actions/#regex

PR #807, opened 2024-06-15
Update .vale.ini

alexvonme:
Links required an update.

### DIFF
--- a/.vale.ini
+++ b/.vale.ini
@@ -8,7 +8,7 @@ SkippedScopes = script, style, pre, figure, code, tt, blockquote, listingblock, 
 
 Packages = RedHat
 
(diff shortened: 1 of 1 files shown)


mmcclaingithub:
@alexvonne Does everyone need to make this update or will running `vale sync`, update the.ini file?

alexvonme:
@mmcclaingithub I think ``vale sync`` [updates the styles folder](https://vale.sh/docs/topics/packages/#style-only) according to the ".vale.ini.". To my knowledge, it won't update the .ini which has to be done manually.

However, my change only updates the sources link from the comments of the ".vale.ini." and as such isn't really required for your setup to continue functioning properly. It really isn't an urgent or breaking change.

aireilly (approved):
/lgtm

PR #809, opened 2024-06-18
Added some phrasal verbs to avoid

gaurav-nelson:
I was going through Applying the IBM Style Guide in Writing Product Documentation and checked to see if out rules had that info covered. 

- Also, we recommend using `accelerate` instead of speed up:
image

### DIFF
--- a/.vale/styles/RedHat/SimpleWords.yml
+++ b/.vale/styles/RedHat/SimpleWords.yml
@@ -12,7 +12,6 @@ swap:
   "objective(?! C?)": aim|goal
   absent: none|not here
   abundance: plenty
-  accelerate: speed up
   accentuate: stress
   accompany: go with
   accomplish: carry out|do
--- a/.vale/fixtures/RedHat/SimpleWords/testinvalid.adoc
+++ b/.vale/fixtures/RedHat/SimpleWords/testinvalid.adoc
@@ -1,6 +1,5 @@
 absent
 abundance
-accelerate
 accentuate
 accompany
 accomplish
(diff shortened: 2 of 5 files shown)


aireilly:
Nice! need a fix here: 

ERROR: 1 in .vale/fixtures/RedHat/SimpleWords/testinvalid.adoc / .vale/styles/RedHat/SimpleWords.yml

aireilly (approved):
/lgtm nice work

chloeredhat:
@aireilly and @gaurav-nelson (@Chandralekha-RedHat for viz) 

Wanted to call out that I think we've made a contradiction lo
Screenshot 2024-06-25 at 10 13 36 AM
op:

.vale/styles/RedHat/SimpleWords.yml
```diff
@@ -12,7 +12,6 @@ swap:
   "objective(?! C?)": aim|goal
   absent: none|not here
   abundance: plenty
-  accelerate: speed up
   accentuate: stress
   accompany: go with
   accomplish: carry out|do
```
chloeredhat, line 17: This is another contradiction. Carry out is a phrasal verb, so accomplish would actually be preferred. Do is a good alternative though 

@aireilly @gaurav-nelson @Chandralekha-RedHat

chloeredhat: Some other thoughts:

plenty: It's better to not put numerical descriptors at all (for example, what does "many applications" mean?) So maybe we shouldn't suggest abundance OR plenty. 

absent: While I agree absent is a complicated word, it is often not going to match grammatically with "none" or "not here". For instance, "If the button is absent" != "*If the button is none" or "*If the button is not here" (not there would make sense in this example)

PR #810, opened 2024-06-19
Add exceptions to the Passive voice rule

aireilly:
```
exceptions:
  - deprecated
  - required
  - imported
  - supported
  - trusted
  - tested
```

~removed~ ~required~

Examples of the above terms in passive usage (lifted from openshift-docs):

1. Service Catalog is deprecated on {product-title} 4 and later.
2. When you create the virtual machine (VM), the data volume with the {object} is imported into persistent storage.
3. This credentials strategy is supported for only new {product-title} clusters and must be configured during installation.
4. Inject the CA bundle that is trusted by {product-title} into the config map by running the following command:
5. The default configuration used by DVM is tested in various environments.

~After the associated credential is removed, it can be deleted or deactivated on the underlying cloud.~

~If you deploy using infrastructure that you manage no additional configuration is required.~

Update: I removed "removed" and "required" from the list since these seems to me to be "primary" verbs that we do want to rewrite in active form for the most part.

### DIFF
--- a/.vale/fixtures/RedHat/PassiveVoice/testvalid.adoc
+++ b/.vale/fixtures/RedHat/PassiveVoice/testvalid.adoc
@@ -1,4 +1,3 @@
-It decided
 It awoken
 It beat
 It become
@@ -31,6 +30,7 @@ It cost
 It crept
 It cut
 It dealt
+It decided
 It dived
 It done
 It drawn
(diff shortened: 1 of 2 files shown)


rolfedh (approved):
LGTM – I fully support this change. Vale's current handling of passive voice generates numerous false or weak notifications, often causing them to be ignored. Refining this rule will reduce unnecessary noise and improve the relevance and accuracy of the notifications.

PR #813, opened 2024-06-21
Small Fix in Contractions.yml

alexvonme:
``aren't`` was corrected to ``is not``. I changed the correction to ``are not``


aireilly (approved):
/lgtm

PR #826, opened 2024-07-03
Fixing Tempo attribute checks

jeana-redhat:
This doesn't fix the RegEx I am trying to nail down, but it does update the Tempo Operator suggestion correctly

More info: https://redhat-internal.slack.com/archives/C0218RXJK5E/p1720025910047249


rohennes (approved):
lgtm

PR #830, opened 2024-07-11
Correct AsciiDoc.ImageContainsAltText criteria

### LINKED ISSUE #829: AsciiDoc.ImageContainsAltText is no longer catching images without alt text
**Describe the bug**
After #737 was merged, AsciiDoc.ImageContainsAltText does not catch images without alt text in asciidoc.

Examples:

* `image::media/marcoverlay5.png[]` no longer creates an AsciiDoc.ImageContainsAltText alert

**To Reproduce**

1. Create a directory with the following basic .vale.ini:
    ```
    StylesPath = vale-styles

    MinAlertLevel = suggestion

    Packages = AsciiDoc

    [*.adoc]
    AsciiDoc.ImageContainsAltText = error
    ```
1. Add this document to your directory as no_alt.adoc:
    ```
    = No alt text =
    
    image::media/marcoverlay1.png[]
    ```
1. Run `vale sync`
2. Run `vale no_alt.adoc`

**Expected behavior**
I would expect the following:

```
 no_alt.adoc
 3:1  error  Image is missing accessibility  AsciiDoc.ImageContainsAltText
             alt tags.

✖ 1 error, 0 warnings and 0 suggestions in 1 file.
```

**Actual behavior**

I get:

```
✔ 0 errors, 0 warnings and 0 suggestions in 1 file.
```

**Additional context**
From my understanding of the [vale docs](https://vale.sh/docs/topics/styles/#existence), if there are multiple elements in a `raw` array, they are simply concatenated together, meaning we are currently looking for the pattern `image::.*\[\]image:.*\[\]`, which won't find much.

I will open a pull request for this issue shortly.

sandbergja:
If there are multiple entries in the raw array, they are concatenated into one large regular expression.  In this case, this meant that we were searching for the regular expression `image::.*\[\]image:.*\[\]`, which doesn't find images without alt text.

In this commit, we use a single regular expression that will match both cases: image with one colon, and image with two colons.

closes #829

### DIFF
--- a/.vale/styles/AsciiDoc/ImageContainsAltText.yml
+++ b/.vale/styles/AsciiDoc/ImageContainsAltText.yml
@@ -5,5 +5,4 @@ level: warning
 link: https://redhat-documentation.github.io/supplementary-style-guide/#cloud-services-images
 message: "Image is missing accessibility alt tags."
 raw:
-  - 'image::.*\[\]'
-  - 'image:.*\[\]'
+  - 'image::?.*\[\]'


aireilly (approved):

aireilly:
Thank you for the PR @sandbergja :partying_face:

PR #834, opened 2024-07-23
Adding PV and PVC

apinnick:
These abbreviations for persistent volume and persistent volume claim are used by OpenShift. See [RHSSG](https://redhat-documentation.github.io/supplementary-style-guide/#persistent-volume).


aireilly:
Should we add PV and PVC to https://github.com/redhat-documentation/vale-at-red-hat/blob/main/.vale/styles/RedHat/Definitions.yml

aireilly (approved):

PR #840, opened 2024-07-24
Adding distro > distribution to Suggestion rule

aireilly:


.vale/styles/RedHat/Spelling.yml
```diff
@@ -49,7 +49,7 @@ filters:
   - "[dD]ev"
   - "[dD]ev[wW]orkspace"
   - "[dD]evfile"
-  - "[dD]evfiles"
```
ccoVeille, line 52: You removed devfiles

Did you do it on purpose?

@aireilly

aireilly: Yes. You don't need to add simple plurals in Spelling rule.

ccoVeille: OK, do you have a documentation you could share about this? I'm interested

PR #848, opened 2024-07-30
Add key/value as exception to Slash rule

apinnick:
I did not find any guidance about using "key/value pairs" in ISG or RHSSG, so I thought it would be good to add as an exception.


apinnick:
Closing because I found something else that might be related. Will reopen if necessary.

PR #852, opened 2024-08-06
Add "re-register" and "re-sign" to hyphens rule

apinnick:
Add "re-register" (error) and "re-sign" (valid) to the hyphens rule.


aireilly:
Push again to retest

aireilly (approved):

PR #878, opened 2024-09-02
Add yaml-patch to case sensitive rule

apinnick:
Add 'yaml-patch' as an exception to the RH case sensitive rule.


aireilly (approved):
/lgtm

PR #879, opened 2024-09-04
Add IBM Secure Execution exception

apinnick:
Add "IBM Secure Execution" as an exception to the OpenShift attributes rule.


aireilly (approved):
/lgtm

PR #882, opened 2024-09-12
Removing IBM suggested attribute rule

aireilly:
The rule that flags "IBM" is too simple and prone to false positives. In general, if a Vale rule is producing false positives, it's not fit for purpose and just adds noise and friction.

"When in doubt, leave it out" :smiley_cat:


apinnick (approved):

dfitzmau:
Change is OK to me. It's very difficult to test these things. I created this rule at the time as I had to crawl the openshift-docs repo to change all instance of IBM names. A rule was a way to flag bad habits of reverting going back to hardcoding IBM names and dropping the RT symbol.

aireilly:
@dfitzmau all the other work you've done there is good though. Kudos. This one is just too generic.

Issue #889, opened 2024-09-30
Rule to flag documentation of self-explanatory fields


apinnick:
Create a rule that flags self-explanatory fields in procedure steps. The purpose of this rule is to reduce 'overdocumentation' of UI procedures.

Example: "Enter your user name in the *Username* field."

Message: "Your procedure mentions %s. Consider not documenting fields that are self-explanatory."

PR #890, opened 2024-09-30
Add ObviousTerms rule

apinnick:
Closes https://github.com/redhat-documentation/vale-at-red-hat/issues/889

### DIFF
--- a/.vale/fixtures/RedHat/ObviousTerms/.vale.ini
+++ b/.vale/fixtures/RedHat/ObviousTerms/.vale.ini
@@ -1,0 +1,5 @@
+; Vale configuration file to test the `ObviousTerms` rule
+StylesPath = ../../../styles
+MinAlertLevel = suggestion
+[*.adoc]
+RedHat.ObviousTerms = YES
--- a/.vale/fixtures/RedHat/ObviousTerms/testinvalid.adoc
+++ b/.vale/fixtures/RedHat/ObviousTerms/testinvalid.adoc
@@ -1,0 +1,2 @@
+Enter your user name in the Username field.
+Enter your description in the Description field.
--- a/.vale/styles/RedHat/ObviousTerms.yml
+++ b/.vale/styles/RedHat/ObviousTerms.yml
@@ -1,0 +1,16 @@
+---
+extends: existence
+level: suggestion
+# link:
+message: "Consider not documenting %s because it is self-explanatory."
+scope: sentence
+# source:
+action:
+  name: remove
+tokens:
+  - 'User field'
+  - 'Username field'
+  - 'Password field'
+  - 'Mail field'
+  - 'Description field'
+  - 'Name field'


.vale/styles/RedHat/ObviousTerms.yml
```diff
@@ -0,0 +1,16 @@
...
+extends: existence
+level: suggestion
+# link:
+message: "Consider not documenting %s because it is self-explanatory."
+scope: list
+# source:
+action:
+  name: remove
+tokens:
+  - User
```
aireilly, line 11: I see where you are going with this, but won't this lead to a lot of false positives?

apinnick: I was thinking about this. We could flag the terms only when bolded (scope: strong?). That should limit occurrences to UI procedures. Or we could flag "**Description** field".

apinnick: Just changed scope to "strong".

.vale/styles/RedHat/ObviousTerms.yml
```diff
@@ -0,0 +1,16 @@
...
+extends: existence
+level: suggestion
+# link:
+message: "Consider not documenting %s because it is self-explanatory."
+scope: raw
+# source:
+action:
+  name: remove
+tokens:
+  - '\*User\*'
```
aireilly, line 11:
```suggestion
  - '\*?User\*? field'
```
Is `**User**` also a valid bold formatting in AsciiDoc?

.vale/fixtures/RedHat/ObviousTerms/.vale.ini
```diff
@@ -0,0 +1,5 @@
+; Vale configuration file to test the `ObviousTerms` rule
+StylesPath = ../../../styles
+MinAlertLevel = suggestion
+[*.adoc]
+RedHat.ObviousTerms = YES
```
aireilly, line 5: I don't like this rule in RedHat because it's highlighting AsciiDoc formatting, and this Vale rule set should be agnostic to formatting.

.vale/fixtures/RedHat/ObviousTerms/testinvalid.adoc
```diff
@@ -0,0 +1,3 @@
+//vale-fixture
```
aireilly, line 1: The Red Hat rule set works by linting line by line. This `//vale-fixture` comment is only required in the rule sets where we don't have one error per line and we need some way to count the errors in the test fixtures.

.vale/styles/RedHat/ObviousTerms.yml
aireilly: Also can we add "field" as extra text in these rules to make it cast iron in terms on only highlighting these fields and not some other valid usage of these terms.

aireilly (commented):
Added some comments, nice work :)

.vale/fixtures/RedHat/ObviousTerms/.vale.ini
apinnick: I agree. Problem is that Vale does not accept "strong" as a scope because it has been deprecated in favor of "raw". Maybe adding "field" is the way to go.

.vale/fixtures/RedHat/ObviousTerms/testinvalid.adoc
apinnick: Ah, OK. So I should remove this comment.

aireilly (approved):
/lgtm

PR #899, opened 2024-10-29
Adding trailing backslash rule

aireilly:
Trailing backslashes in codeblocks break the openshift-docs Asciibinder build. Backslashes are only a problem in asciibinder when they come just before the end of a codeblock.

### DIFF
--- a/.vale/fixtures/OpenShiftAsciiDoc/TrailingBackslash/testvalid.adoc
+++ b/.vale/fixtures/OpenShiftAsciiDoc/TrailingBackslash/testvalid.adoc
@@ -1,0 +1,46 @@
+[source,terminal]
+----
+$ oc get my lunch
+----
+
+[source,terminal]
+----
+# oc get my lunch
+----
+
+.Example
+[source,terminal]
+----
+$ az role assignment create --role "<privileged_role>" \// <1>
+----
+
+[source,terminal]
+----
+sh-4.2# chroot /host
+----
+
+[source,terminal]
+----
+.
+├── CODEOWNERS
+├── README.md
+└── scripts
+    └── validate-vale-rules.sh
+----
+
+[source,terminal]
+----
+(undercloud)$ cd go/to/path
+----
+
+[source,yaml]
+----
+include::yaml/boundary-clock-ptp-config.yaml[]
+----
+
+.Example
+[source,terminal]
+----
+$ az role assignment \
+  create --role "<privileged_role>"
+----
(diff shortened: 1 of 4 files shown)


.vale/fixtures/OpenShiftAsciiDoc/TrailingBackslash/testvalid.adoc
```diff
@@ -0,0 +1,39 @@
...
+
+[source,terminal]
+----
+# oc get my lunch
+----
+
+.Example
+[source,terminal]
+----
+$ az role assignment create --role "<privileged_role>" \// <1>
```
apinnick, line 14: Could you add a "normal" backslash without a callout to the testvalid file?

----
$ az z role assignment \
  create --role "<privileged_role>"
----

aireilly: This will trigger the rule. I'll see if I can refine the regex to cater for this.

apinnick: I ran into a similar issue when I tried to create a Vale rule that would flag lines in code that were missing backslashes. :-(  Never did find a good solution.

aireilly: Could handle it with a script rule for sure.

aireilly: Fixed this :sparkles: PTAL now

apinnick: I'm not sure how you did this but it seems to have passed the check. LGTM

.vale/styles/OpenShiftAsciiDoc/TrailingBackslash.yml
```diff
@@ -0,0 +1,9 @@
+---
+extends: existence
+scope: raw
+level: error
+message: "Code example ends with an unescaped trailing backslash."
+action:
+  name: remove
+raw:
+ - '\\(?=\n----)'
```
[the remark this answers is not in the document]
aireilly: Yep. Added :)

.vale/fixtures/OpenShiftAsciiDoc/TrailingBackslash/testvalid.adoc
aireilly: Changed the regex to work "backwards", since it's only a newline that comes directly before the closing codeblock delimiter that causes the problem.

PR #901, opened 2024-11-26
Warn when 'police' is used in place of 'policy'

jhradilek:
Because of the similarity of the two words, it is [not uncommon](https://docs.redhat.com/search/?p=1&rows=10&documentKind=Documentation&q=polices) to find the word 'police' or 'polices' used in place of 'policy' or 'policies'. As the former is highly unlikely to appear in software documentation, I propose to issue a warning when this word is found.


aireilly (approved):
/lgtm thanks for the PR!

PR #905, opened 2025-01-08
Add 'k8s' to TermsErrors

apinnick:
Add 'k8s' to TermsErrors.
Removed 'k8s' from CaseSensitiveTerms because the rule does not find 'K8s'.

### DIFF
--- a/.vale/styles/RedHat/TermsErrors.yml
+++ b/.vale/styles/RedHat/TermsErrors.yml
@@ -233,6 +233,7 @@ swap:
   joblog: job log
   jobstream: job stream
   judgement: judgment
+  k8s: Kubernetes
   keep in mind: remember
   kernelspace: kernel-space
   kick off: start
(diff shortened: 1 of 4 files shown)


.vale/styles/RedHat/TermsErrors.yml
```diff
@@ -233,6 +233,7 @@ swap:
   joblog: job log
   jobstream: job stream
   judgement: judgment
+  k8s: Kubernetes
```
aireilly, line 236: Wondering. Do we have legitimate instances of things like namespaces etc with k8s in the name? Maybe with dashes? Does this rule trigger in those instances?

aireilly (approved):
/lgtm

.vale/styles/RedHat/TermsErrors.yml
apinnick: Update: When I removed the occurrences that were detected in auto-generated API docs, the results were 7 occurrences in 4 files.

PR #915, opened 2025-02-17
kube-burner is a valid term

aireilly:


.vale/fixtures/RedHat/TermsErrors/testvalid.adoc
```diff
@@ -358,6 +358,7 @@ kernel oops
 kernel-space
 keyboard shortcut
 knowledge base
+kube-burner
```
apinnick, line 361: It's the name of a [tool](https://kube-burner.github.io/kube-burner/latest/), so I think it should be capitalized.

apinnick (changes requested):

.vale/fixtures/RedHat/TermsErrors/testvalid.adoc
aireilly: The terms errors rule as-is will catch "Kube" and "kube".

`ignorecase: true`

PR #918, opened 2025-02-17
Only error on k8s when it is a lone word

aireilly:


apinnick (approved):

apinnick:
This might be a problem. I recall a writer recently using 'OVN-k8s'.

Issue #921, opened 2025-03-06
Consider removing "address" from "use simple language" rule


jeana-redhat:
**Is your feature request related to a problem? Describe.**
In tech docs, we often discuss "addresses" (as a noun or networking concept). Currently, Vale warns about the string `address` and suggests using a more simple verb. 

A brief Ctrl+Shift+F based tour of the OpenShift docs repo is showing at least 60% of uses of "address" are referring to computer concepts rather than a discussion, as Vale suggests. 

The more false positives we have, the less likely we are to notice and address (😉) actual issues with word use.

**Describe the solution you'd propose**
Consider removing "address" from "use simple language" rule. 

It might make sense to do some research to see if it's more of a false-positive than real hit, but I don't know if that's necessary or if we even have a rule of thumb to apply to any data we collect.

**Describe alternatives you've considered**
* Ignore it and leave it as it is.
* Make a more complex rule that can tell if you're talking about an IP address or similar.

**Additional context**
[Slack discussion](https://redhat-internal.slack.com/archives/C0218RXJK5E/p1741185415516489)
Example:
![Image](https://github.com/user-attachments/assets/67ac847a-9a7c-4790-8d1a-4f9e8ce0f10a)

apinnick:
I think removing "address" is a good idea.

PR #922, opened 2025-03-07
Tweak address

aireilly:
Adjusts rule based on https://github.com/redhat-documentation/vale-at-red-hat/issues/921

### DIFF
--- a/.vale/fixtures/RedHat/SimpleWords/testinvalid.adoc
+++ b/.vale/fixtures/RedHat/SimpleWords/testinvalid.adoc
@@ -9,7 +9,6 @@ accrue
 accurate
 acquiesce
 acquire
-address
 addressees
 adjacent to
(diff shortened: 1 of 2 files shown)


aireilly:
> Can you explain this change to me? Why replace with "discuss"?

Hmm. Can't find anything to back this up, and the link in the rule includes the verb "addressing" :rofl: 

https://www.ibm.com/docs/en/ibm-style?topic=medium-conversational-style

I think probably the word might be less than ideal from a localization perspective? 

Note - I did not add this rule, just updating it with this PR.

apinnick:
Oh, now I see -- you want to avoid "address" as a verb. Honestly, I'm not sure this rule is a good idea. It's very likely to result in false positives.

aireilly:
Removing this rule!

PR #933, opened 2025-04-22
Tweak k8s in TermsErrors.yml

aireilly:
k8s should not raise in error in API names.

### DIFF
--- a/.vale/styles/RedHat/Spelling.yml
+++ b/.vale/styles/RedHat/Spelling.yml
@@ -437,6 +437,7 @@ filters:
   - precache
   - Prometheus
   - proxied
+  - PVCs
   - Pytorch
   - qdmanage
   - qdstat


.vale/styles/RedHat/TermsErrors.yml
```diff
@@ -233,7 +233,7 @@ swap:
   joblog: job log
   jobstream: job stream
   judgement: judgment
-  '(?<!-)\bk8s\b(?!-)': Kubernetes
+  '(?<!-|\.)\bk8s\b(?!-|\.)': Kubernetes
```
apinnick, line 236: Is it not possible to set the scope to ignore code?

aireilly: Sometimes API names are not marked in code blocks, and this term is so pervasive I would be inclined to remove it entirely from the errors list. Remember, errors can cause a CI to fail, and we only want cast iron errors in this list.

In fact, I'd probably be inclined to move this to warnings or suggestions.

aireilly: WDYT @apinnick?

apinnick: I agree. Remove it or downgrade it. I think there is too great a risk of false positives for this term.

A lot of OpenShift code comes from upstream projects. Maintaining this particular 'error' could turn into a nightmare.

PR #954, opened 2025-07-21
Update regex to match ifeval only at the beginning of a line

### LINKED ISSUE #953: ValidConditions should not detect `ifeval` directives in comments
ValidConditions.yml detects balanced `ifeval` directives as _unbalanced_ when they are commented out, because the rule matches `ifeval` but not `endif` in comments.

https://github.com/redhat-documentation/vale-at-red-hat/blob/792218023e6786d88ec2eb40485f6ee12955c9df/.vale/styles/AsciiDoc/ValidConditions.yml#L16C1-L19C36

### Problem
In line 18, `ifeval_regex := "ifeval::\\[.+\\]"` matches directives in comments.
### Solution
Change to `ifeval_regex := "^ifeval::\\[.+\\]"`

ryan-carpenter:
Fix #953.

### DIFF
--- a/.vale/styles/AsciiDoc/ValidConditions.yml
+++ b/.vale/styles/AsciiDoc/ValidConditions.yml
@@ -15,7 +15,7 @@ script: |
 
   if_regex := "^ifdef::.+\\[\\]"
   ifn_regex := "^ifndef::.+\\[\\]"
-  ifeval_regex := "ifeval::\\[.+\\]"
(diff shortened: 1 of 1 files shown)


ryan-carpenter:
I've updated the Tengo script but haven't run it yet, since I don't have Tengo installed. I'll need to put that off for a couple of days, but can install and properly validate then if you would like. I did "validate" the yaml file with a regex search in the editor.

PR #955, opened 2025-08-09
fix: allow the transitive phrasal verb "built in"

edcsnt:
"built in" should not be unconditionally substituted by "built-in". "Built in" is the valid past principle form of the transitive phrasal verb "to build in", as [seen](https://www.oed.com/dictionary/build_v?tab=phrasal_verbs#1351285480) in the Oxford English Dictionary.


aireilly (approved):
/lgtm

PR #987, opened 2025-11-24
OpenShiftAsciiDoc - no alphanumeric charaters in headings

rohennes:
PV1 fails late in the build process if the headings aren't alphanumeric. It balks at even having periods in them.


.vale/fixtures/OpenShiftAsciiDoc/HeadingsAlphanumeric/testinvalid.adoc
```diff
@@ -0,0 +1,8 @@
+//vale-fixture
+= Don't Use: Special Characters in Headings
+
+//vale-fixture
+== No @Symbols! Here
+
+//vale-fixture
+== No periods.
```
aireilly: Yes but Something like this is valid and gets caught too:

== Installing the .NET plugin

PR #991, opened 2025-12-23
Request to add CNV common terms to Vale

abrennan89:
Some terms are repeatedly flagged as warnings for CNV docs, which means that technically they're not passing the Vale check required as part of our CQAs.

I am proposing adding some commonly used terms to the spellchecker dictionary so that CNV docs can pass. I couldn't find any guidance anywhere forbidding the use of these terms.


aireilly:
Is this ready to merge?

PR #1008, opened 2026-02-04
Adding no gerunds in titles Suggestion rule

aireilly:
Red Hat docs guidance is now _not_ to use gerunds in titles, in favor of the imperative form.

This suggestion rule uses a tengo script replace action to suggest a correct replacement.

image


jeana-redhat (commented):
Oops I thought I hit submit on my comment but I guess it started a review

aireilly:
lol forgot about this but I guess we can push the button now?

PR #1034, opened 2026-05-14
Remove suggestions for the unsupported MTC attributes

jhradilek:
As stated in [MIG-1877](https://redhat.atlassian.net/browse/MIG-1877), Migration Toolkit for Containers is only supported in OpenShift up to version 4.20. Consequently, [pull request #108725](https://github.com/openshift/openshift-docs/pull/108725) removed the attributes form the corresponding attribute definition files and any references to them from `main`, `enterprise-4.21`, and `enterprise-4.22`. Because of this, I believe the `OpenShiftAsciiDoc.SuggestAttribute` rule should no longer suggest the use of these removed attributes.


jhradilek:
@aireilly Hi, I am not sure who maintains this specific style. I am also not sure if the failing preview job is something I can influence, it does not look like I can from the error message.

aireilly:
No worries /lgtm

PR #1047, opened 2026-07-16
Include config/actions in RedHat.zip release

aireilly:
## Summary
- The `NoGerundsInTitles` rule references a tengo script (`NoGerundsInTitles.tengo`) via its `action.suggest` config, but the script lives in `config/actions/` which was not included in the `RedHat.zip` release artifact.
- Users who `vale sync` the RedHat package get `script 'NoGerundsInTitles.tengo' not found` because the zip only contained the `RedHat/` directory.
- Adds `config/actions` to the `zip` command so the tengo script is bundled in the release.

## Test plan
- [ ] Verify the release workflow produces a `RedHat.zip` that contains both `RedHat/` and `config/actions/NoGerundsInTitles.tengo`
- [ ] Run `vale sync` with the new zip and confirm `NoGerundsInTitles` rule works without errors

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -44,7 +44,13 @@ jobs:
         run: |
           cd .vale/styles
           echo $GITHUB_RUN_NUMBER > version.txt
-          zip -r RedHat.zip RedHat config
+          mkdir -p /tmp/redhat-pkg/styles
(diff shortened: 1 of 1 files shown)


aireilly:
Closing in favour of https://github.com/redhat-documentation/vale-at-red-hat/pull/1046

